PR #25, opened 2026-02-08
Feat/anonymous occupancy events

chuckmilton:
### Overview
Implements a privacy-first anonymous event logging system for parking lot ENTER/EXIT detection. Events from mobile geofencing are stored in DynamoDB with device IDs hashed server-side for deduplication without storing PII. A 15-minute cron job generates occupancy snapshots for ML forecasting model training.

### Features

**Backend (OccupancyEventsModule)**
- New `/api/v1/occupancy-events` endpoint for recording anonymous ENTER/EXIT events
- SHA-256 device ID hashing with configurable salt for privacy-safe deduplication
- Automatic duplicate detection (prevents ENTER→ENTER or EXIT→EXIT sequences)
- Atomic occupancy counter updates using DynamoDB conditional writes
- 15-minute scheduled job (`@nestjs/schedule`) for ML training snapshot generation
- Event statistics endpoints for debugging and analysis
- 90-day TTL for automatic data expiration (GDPR compliance)

**Mobile Integration**
- `@react-native-async-storage/async-storage` for persistent anonymous device ID
- Updated `recordOccupancyEvent()` to include device_id in payload
- Fallback to session-only ID if AsyncStorage is unavailable

**DynamoDB Schema Updates**
- New GSI (`GSI1-LotID-Timestamp`) on timeseries table for efficient queries
- Device deduplication tracking records

### API Endpoints
| Method | Endpoint | Description |
|--------|----------|-------------|
| POST | `/occupancy-events` | Record ENTER/EXIT event |
| GET | `/occupancy-events/lots/:lotId` | Get events by lot (with date range) |
| GET | `/occupancy-events/lots/:lotId/stats` | Get enter/exit counts |
| POST | `/occupancy-events/snapshots` | Manually trigger snapshot creation |
| GET | `/occupancy-events/snapshots/:lotId` | Get snapshots by date |

### Files Changed

**Backend - New Files (11)**
- `src/occupancy-events/occupancy-events.module.ts`
- `src/occupancy-events/occupancy-events.service.ts` + `.spec.ts`
- `src/occupancy-events/occupancy-events.controller.ts` + `.spec.ts`
- `src/occupancy-events/occupancy-events.scheduler.ts` + `.spec.ts`
- `src/occupancy-events/dto/create-occupancy-event.dto.ts`
- `src/occupancy-events/interfaces/occupancy-event.interface.ts`
- `src/occupancy-events/utils/privacy.util.ts` + `.spec.ts`
- `test/occupancy-events.e2e-spec.ts`

**Backend - Modified Files**
- `app.module.ts` - Import OccupancyEventsModule
- package.json - Added `@nestjs/schedule`
- `lots.service.ts` / `lots.controller.ts` - Removed deprecated recordOccupancyEvent

**Mobile - Modified Files**
- `src/services/api/lots.ts` - Updated with device_id and AsyncStorage
- `src/services/api/config.ts` - New OCCUPANCY_EVENTS endpoint
- package.json - Added `@react-native-async-storage/async-storage`
- Updated test files for new response format

**Scripts**
- `setup-dynamodb-schema.ts` - Added GSI for lot/timestamp queries
- `seed-database.ts` - Added `seedOccupancyEvents()` for sample data

### Test Coverage
- **46 unit tests** covering service, controller, scheduler, and privacy utilities
- **16 e2e tests** for all API endpoints including validation and edge cases
- All existing tests pass

### Privacy Considerations
- Device IDs are hashed with SHA-256 + salt before storage
- No user coordinates or PII stored in events
- TTL ensures automatic deletion after 90 days
- Device hash only used for deduplication, not identification

### DIFF
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -20,6 +20,7 @@
     "@nestjs/core": "^11.0.1",
     "@nestjs/passport": "^11.0.5",
     "@nestjs/platform-express": "^11.0.1",
+    "@nestjs/schedule": "^6.1.1",
     "@nestjs/swagger": "^11.2.3",
     "class-transformer": "^0.5.1",
     "class-validator": "^0.14.3",
--- a/apps/backend/src/app.module.ts
+++ b/apps/backend/src/app.module.ts
@@ -7,6 +7,7 @@ import { UsersModule } from './users/users.module';
 import { EventsModule } from './events/events.module';
 import { WeatherModule } from './weather/weather.module';
 import { AuthModule } from './auth/auth.module';
+import { OccupancyEventsModule } from './occupancy-events/occupancy-events.module';
 
 @Module({
   imports: [
@@ -15,7 +16,8 @@ import { AuthModule } from './auth/auth.module';
     UsersModule,
     EventsModule,
     WeatherModule,
-    AuthModule
+    AuthModule,
+    OccupancyEventsModule,
   ],
   controllers: [AppController],
   providers: [AppService],
--- a/apps/backend/src/lots/lots.controller.spec.ts
+++ b/apps/backend/src/lots/lots.controller.spec.ts
@@ -1,5 +1,4 @@
 import { Test, TestingModule } from '@nestjs/testing';
-import { BadRequestException } from '@nestjs/common';
 import { LotsController } from './lots.controller';
 import { LotsService } from './lots.service';
 
@@ -12,7 +11,6 @@ describe('LotsController', () => {
     findOne: jest.fn(),
     getHistory: jest.fn(),
     getOccupancySummary: jest.fn(),
-    recordOccupancyEvent: jest.fn(),
   };
 
   beforeEach(async () => {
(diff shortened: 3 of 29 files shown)


DottL (approved):
Great work on the scheduler!! 
Capturing continuous snapshots allows the model to learn the lots' occupancy velocity, which should also improve our peak-time prediction accuracy.

Image

lawrenceDegoma (approved):
Good work on the occupancy events, i'll get to work connecting our ends together. #Shoutout_SHA_256

On a side note though, are there any other metrics we want to keep track of? Like user speeds and whatnot that I'll need to track anyways for parking verification?

chuckmilton:
> Good work on the occupancy events, i'll get to work connecting our ends together. #Shoutout_SHA_256
> 
> On a side note though, are there any other metrics we want to keep track of? Like user speeds and whatnot that I'll need to track anyways for parking verification?

Thank you @lawrenceDegoma for the detailed review! For parking verification, I would say that there are a few conditions we can keep track of.

**Entry Detection / Parking Verification**:
- When a user enters a parking lot, there are some conditions we can consider to verify if a user is officially parked at a parking lot.
  - If the user was connected to the car via BlueTooth, Apple CarPlay, etc. upon entering a parking lot, and they disconnect from it (which usually happens by unplugging their phone, turning their car off, etc.) that is a good indicator that they most likely parked.
  - Another factor would be the significant decrease in speed when entering into the lot. For instance, if the user was going 25mph and it slows down to 0-5mph upon entering the parking lot, that would be a good indicator. 
    - Or if the user is staying at one spot in the lot for awhile and its at 0mph, thats a good indicator too. 
    - Another one could be if the user leaves the parking lot at a significantly slower speed compared to leaving a parking lot with a car, theres a good chance that they parked.

**Leave Detection / Leave Verification**:
- Pretty much the same thing but vice-versa. We could keep track of other things tho
  - If the user comes back to the same parking lot that we classified they were parked in earlier, it could be a good chance that they're about to leave soon.
  - If they connect to their car via bluetooth, Apple CarPlay, etc. then they most likely have their car turned on and are most likely about to leave soon.
  - A big factor would be if they leave the parking lot and the speed significantly increases, its a good chance they left that parking lot with their car.
    - If they leave the lot and the speed increases and they don't come back for awhile, they most likely left
    - The most determining factor would be if they leave campus in general, we can be almost absolutely sure they left.

Overall, the main metrics you'd be keeping track of will be the user's speed, whether their phone is connected to their car or not, their parking state (remembering if they're parked or not), and their dwell time. We could definitely add more metrics for better accuracy and consider certain use cases, but this is the general idea. All of this will be kept track from the client side, since our goal is to store the least amount of data we need from our users.

But yeah, feel free to make any changes/improvements in this logic. These are just ideas for now, but hopefully this gives you a better picture of how we should approach the enter/leave detection.

PR #29, opened 2026-02-21
Feat/favorite lots

zachfpadilla:
## Summary
Implements the ability for logged-in users to favorite & unfavorite any given lot. In addition, the navigation FAB of the map screen displays two speed dial options when pressed: "recommended" & "favorites". 

## Additions
`/services/api/favorites.ts`
- Service for invoking favorites-related endpoints (`GET :userId/favorites` for retrievel, `<POST|DELETE> :userId/favorites/:lotId` for addition/removal).
- If user credentials provided for a call are expired, an attempt will be made to refresh the session & the endpoint will be tried again.
- Tests have been created for each of these cases.

`/hooks/useFavorites.ts`
- Hook for utilizing the services provided by the favorites API. Sends & receives user credentials between each end.
- Manages a local copy of the user's favorited lots to ensure that components immediately reflect successful changes.
- Tests have been created for component rendering & to ensure accurate interfacing.

`/components/NavigationSpeedDial.tsx`
- This is composed of the two options that appear after pressing the navigation FAB of the main screen: "recommended" & "favorites".
- Either option will display a navigation modal with the appropriate set of lots ("recommended" currently defaults to all registered lots).

`/components/Modals/NavigationModal.tsx`
- Menu that displays a list of appropriate lots with names & estimated occupancies included.
- When a lot is selected, the user is redirected to the appropriate short-term forecast screen.

## Modifications
`/auth/AzureAuth.tsx`
- Configured an `access_as_user` API via Microsoft Entra and added it to the scope config.
   - This is needed for users to be able to make API calls using their access tokens.
- `userId` is now an attribute of `AuthResult`, as it acts as the identifier of the current user.

`/context/AuthContext.tsx`
- Implemented & exported `refreshSession` so that session refreshing can be performed by other processes, e.g. when attempting to access user endpoints.

`/screens/MapScreen.tsx`
- Added the navigation modal & the values required for it to be functional (e.g. the lots to display).
- The navigation FAB is now passed to & wrapped within the navigation speed dial options, as the two are loosely coupled.

`/screens/ShortTermForecastScreen.tsx`
- Implemented a button for toggling the favorited status of the current lot. Changing its status reflects immediately both locally  & in the backend.
- An alert will be displayed if the backend cannot be communicated with, and any local changes are nullified.

`/services/api/base.ts`
- Added an optional `options` parameter to `get`, `post`, `put`, and `delete`. This is necessary in order to pass authorization/bearer information to any of these requests when needed, e.g. for invoking user endpoints that require an access token.

## Screenshots
image
image
image

### DIFF
--- a/apps/mobile/__tests__/useFavorites.test.ts
+++ b/apps/mobile/__tests__/useFavorites.test.ts
@@ -1,0 +1,118 @@
+/**
+ * Tests for useFavorite hooks
+ */
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import { useFavorites, UseFavoritesReturn } from '../src/hooks/useFavorites';
+import { useAuth } from '../src/context/AuthContext';
+import { AuthResult } from '../src/auth/AzureAuth';
+import favoritesApi from '../src/services/api/favorites';
+
+jest.mock('../src/context/AuthContext');
+jest.mock('../src/services/api/favorites');
+
+const mockUseAuth = useAuth as jest.MockedFunction<typeof useAuth>;
+const mockFavoritesApi = favoritesApi as jest.Mocked<typeof favoritesApi>;
+
+/**
+ * Harness to test hook functionality, due to the lack of a testing library
+ */
+function HookTestComponent({ onHookValue }: { onHookValue: (val: UseFavoritesReturn) => void }) {
+  const hookVal = useFavorites();
+  onHookValue(hookVal);
+  return null;
+}
+
+describe('useFavorites hooks', () => {
+  const mockUser: AuthResult = {
+    accessToken: 'mock-access-token',
+    idToken: 'mock-id-token',
+    refreshToken: 'mock-refresh-token',
+    accessTokenExpirationDate: new Date().toISOString(),
+    tokenType: 'Bearer',
+    userId: 'test-user-id',
+  };
+
+  const mockRefresh = jest.fn();
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockUseAuth.mockReturnValue({
+      user: mockUser,
+      refreshSession: mockRefresh,
+      isAuthenticated: true,
+      login: jest.fn(),
+      logout: jest.fn(),
+      isLoading: false,
+    });
+  });
+
+  describe('refreshFavorites', () => {
+    it('fetches favorite lots on component mount', async () => {
+      const mockLots = ['G1', 'G3'];
+      mockFavoritesApi.getAllFavorites.mockResolvedValueOnce(mockLots);
+
+      let hookResult: UseFavoritesReturn;
+      // Manually capture hook state
+      const capture = (val: UseFavoritesReturn) => { hookResult = val; };
+
+      // Simulating the effect trigger
+      await (async () => {
+        await favoritesApi.getAllFavorites(mockUser.userId, mockUser.accessToken, mockRefresh);
+      })();
+
+      expect(mockFavoritesApi.getAllFavorites).toHaveBeenCalledWith(
+        mockUser.userId,
+        mockUser.accessToken,
+        mockRefresh
+      );
+    });
+  });
+
+  describe('Optimistic updates', () => {
+    it('handles optimistic updates correctly', async () => {
+      const newLot = 'G5';
+      mockFavoritesApi.addFavorite.mockResolvedValueOnce(undefined);
+
+      const state: string[] = ['G1'];
+      const optimisticState = [...state, newLot];
+
+      expect(optimisticState).toContain(newLot);
+
+      await favoritesApi.addFavorite(mockUser.userId, mockUser.accessToken, newLot, mockRefresh);
+      expect(mockFavoritesApi.addFavorite).toHaveBeenCalled();
+    });
+
+    it('rollbacks state if the API fails', async () => {
+      const failedLot = 'G10';
+      mockFavoritesApi.addFavorite.mockRejectedValueOnce(new Error('API Fail'));
+
+      let favoriteLots: string[] = [];
+      try {
+        favoriteLots = [failedLot];
+        await favoritesApi.addFavorite(mockUser.userId, mockUser.accessToken, failedLot, mockRefresh);
+      } catch (e) {
+        favoriteLots = favoriteLots.filter(id => id !== failedLot);
+      }
+
+      expect(favoriteLots).not.toContain(failedLot);
+    });
+  });
+
+  describe('AuthGuard verification', () => {
+    it('does not perform operations if user is null', async () => {
+      mockUseAuth.mockReturnValue({
+        user: null,
+        refreshSession: mockRefresh,
+        isAuthenticated: false,
+        login: jest.fn(),
+        logout: jest.fn(),
+        isLoading: false,
+      });
+
+      const userId = null;
+      if (!userId) {
+         expect(mockFavoritesApi.getAllFavorites).not.toHaveBeenCalled();
+      }
+    });
+  });
+});
--- a/apps/mobile/__tests__/useFavorites.test.ts
+++ b/apps/mobile/__tests__/useFavorites.test.ts
@@ -2,7 +2,8 @@
  * Tests for useFavorite hooks
  */
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import { useFavorites, UseFavoritesReturn } from '../src/hooks/useFavorites';
+import { renderHook, act, waitFor } from '@testing-library/react-native';
+import { useFavorites } from '../src/hooks/useFavorites';
 import { useAuth } from '../src/context/AuthContext';
 import { AuthResult } from '../src/auth/AzureAuth';
 import favoritesApi from '../src/services/api/favorites';
@@ -12,15 +13,6 @@ jest.mock('../src/services/api/favorites');
 
 const mockUseAuth = useAuth as jest.MockedFunction<typeof useAuth>;
 const mockFavoritesApi = favoritesApi as jest.Mocked<typeof favoritesApi>;
-
-/**
- * Harness to test hook functionality, due to the lack of a testing library
- */
-function HookTestComponent({ onHookValue }: { onHookValue: (val: UseFavoritesReturn) => void }) {
-  const hookVal = useFavorites();
-  onHookValue(hookVal);
-  return null;
-}
 
 describe('useFavorites hooks', () => {
   const mockUser: AuthResult = {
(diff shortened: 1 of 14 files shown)


apps/mobile/src/hooks/useFavorites.ts
```diff
@@ -0,0 +1,101 @@
...
+  const [favoriteLots, setFavoriteLots] = useState<string[]>([]);
+  const [isLoading, setLoading] = useState<boolean>(true);
+  const [error, setError] = useState<string | null>(null);
+
+  const userId = user?.userId;
+  const accessToken = user?.accessToken;
+
+  useEffect(() => {
+    refreshFavorites();
+  }, [userId, accessToken]);
```
chuckmilton: Minor fix: The `eslint-plugin-react-hooks` exhaustive-deps rule would flag this because `refreshFavorites` is used inside the effect but not listed as a dependency. In practice this doesn't cause a runtime bug here since both the effect and the `useCallback` share the same deps (userId, accessToken), so they stay in sync. But it's still a lint violation and could become a real issue if the deps diverge in the future. It'd be best if we either add `refreshFavorites` to the array or call the API inline inside the effect.

apps/mobile/src/screens/ShortTermForecastScreen.tsx
```diff
@@ -90,10 +104,33 @@ export function ShortTermForecastScreen() {
...
+      Alert.alert(
+        'Favorite Lots Error',
+        'The favorite status of the lot could not be changed. Please try again.',
+        [{ text: 'OK', style: 'cancel' }]
+      );
+    }
+  };
+
+  // Favorite button component
+  const FavoriteButton: React.FC<{ isFavorite: boolean; onToggle: () => void }> = ({ isFavorite, onToggle }) => (
```
chuckmilton: `FavoriteButton` is declared as a React component inside `ShortTermForecastScreen`'s function body (after the early returns). This means React creates a brand-new component type on every render. Instead of reconciling prop changes, React will unmount and remount the button each time, which can cause flicker, lost focus state, and unnecessary work. What would be best is to move it outside the component.

apps/mobile/src/components/Modals/NavigationModal.tsx
```diff
@@ -0,0 +1,205 @@
...
+              <ActivityIndicator size="large" color={COLORS.primary} />
+            </View>
+          ) : displayLots.length === 0 ? (
+                /* Empty State View */
+                <View style={styles.emptyContainer}>
+                  <Text style={styles.emptyTitle}>No Favorites Yet</Text>
+                  <Text style={styles.emptySubtext}>
+                    Favorited lots will appear here for quick navigation.
+                  </Text>
+                </View>
```
chuckmilton: The empty state always says "No Favorites Yet" / "Favorited lots will appear here for quick navigation." even when the user tapped the "Recommended" button. Since this modal is used for both favorites and recommended lots, the copy should change based on which mode opened it.

I would add a mode prop (or emptyTitle/emptySubtext props) so the caller can control what's shown

apps/mobile/__tests__/useFavorites.test.ts
```diff
@@ -0,0 +1,118 @@
...
+        logout: jest.fn(),
+        isLoading: false,
+      });
+
+      const userId = null;
+      if (!userId) {
+         expect(mockFavoritesApi.getAllFavorites).not.toHaveBeenCalled();
+      }
+    });
+  });
```
chuckmilton, line 162: A `HookTestComponent` is defined on line 20 but never rendered anywhere. Every test in this file calls the mocked API directly or manipulates plain arrays manually, rather than mounting the hook and testing its actual state transitions.

For example, the "fetches favorite lots on component mount" test (line 51) does this:

```typescript
await favoritesApi.getAllFavorites(mockUser.userId, mockUser.accessToken, mockRefresh);
```
That's just calling the mock yourself, but it doesn't test that the hook's `useEffect` fires, that `favoriteLots` state gets set, or that `isLoading` transitions correctly.

Similarly, the optimistic update tests (lines 71–100) simulate state changes with local variables:
```typescript
const state: string[] = ['G1'];
const optimisticState = [...state, newLot];
```

This doesn't test that calling `addFavorite` from the hook actually triggers the optimistic update and rollback logic.

And the auth guard test (lines 105–117) creates a local null variable and checks it, the hook is never invoked, so the assertion passes trivially.

For this, I would probably consider using `renderHook` from `@testing-library/react-native` (or `@testing-library/react-hooks`) to actually mount the hook and test real behavior:

```typescript
const { result, waitForNextUpdate } = renderHook(() => useFavorites(), { wrapper });
await waitForNextUpdate();
expect(result.current.favoriteLots).toEqual(['G1', 'G3']);
```

apps/mobile/src/services/api/favorites.ts
```diff
@@ -0,0 +1,82 @@
...
+   */
+  private async requestWithRetry<T>(
+    apiCall: (token: string) => Promise<T>,
+    accessToken: string,
+    refreshSession: () => Promise<AuthResult | null>
+  ): Promise<T> {
+    try {
+      return await apiCall(accessToken);
+    } catch (error) {
+      if ((error as ApiError).status == 401) {
```
chuckmilton: We should use strict equality (===). It works since both sides are numbers, but it should be consistent with TypeScript best practices and the rest of the codebase.

apps/mobile/src/screens/MapScreen.tsx
```diff
@@ -158,29 +164,48 @@ const MapScreen: React.FC = () => {
...
+  const openNavigationModal = async (type: 'favorites' | 'recommended') => {
+    if (type === 'favorites') {
+      const favoriteLots = await refreshFavorites(); // Ensures that current instance of favoriteLots is up-to-date
+      setNavigableLots(favoriteLots);
+    } else {
+      // TODO: Implement lot recommendation system; using all lots as a placeholder
+      setNavigableLots(parkingLots.map(lot => lot.id));
+    }
+    setIsNavigationModalOpen(true);
+  };
```
chuckmilton: From what I understand, the navigate button now opens a speed dial → modal → lot list → user picks → navigates to forecast. This is essentially a browse-and-pick flow.

Was the original intent for the nav button to automatically pick the best lot and navigate directly (i.e., the app decides for you), rather than presenting a list for the user to choose from? If so, the current implementation changes the expected UX from "navigate me" to "show me options." The "Recommended" option already has a TODO for actual recommendation logic (which is completely fine), just want to make sure the team is aligned on whether the interaction model here should be "auto-select" or "user-picks-from-list," since that's a bigger design decision than the placeholder logic.

apps/mobile/src/hooks/useFavorites.ts
```diff
@@ -0,0 +1,101 @@
...
+      // Rollback on failure
+      setFavoriteLots(previousFavorites);
+
+      const errorMessage = error instanceof ApiError
+        ? `${error.message} (${error.status})`
+        : `[useFavorites] Failed to unfavorite lot ${lotId} for ${userId}`;
+      setError(errorMessage);
+      throw error;
+    }
+  }, [favoriteLots, userId, accessToken]);
```
chuckmilton: Including `favoriteLots` in the `useCallback` deps means a new `removeFavorite` function reference is created every time the favorites array changes. Any component that depends on `removeFavorite` identity (for example, in its own `useEffect` or `useMemo`) will re-run unnecessarily.

The `favoriteLots` dependency is there because of `const previousFavorites = [...favoriteLots]` on line 83 (for rollback). You could avoid this by using a ref:
```typescript
const favoritesRef = useRef(favoriteLots);
favoritesRef.current = favoriteLots;

const removeFavorite = useCallback(async (lotId: string) => {
  const previousFavorites = [...favoritesRef.current];
  // ...
}, [userId, accessToken]);
```

Just a minor optimization, not blocking 🙏

chuckmilton (changes requested):
Amazing and solid work overall! The favorites feature is clean and well-structured. I left a few minor suggestions on some of the files, nothing major. The optimistic update pattern is a nice touch as well. But yeah, I think after these changes, everything should be good to go. Great work 🙌

apps/mobile/src/hooks/useFavorites.ts
```diff
@@ -0,0 +1,101 @@
...
+      setFavoriteLots(favorites);
+      return favorites;
+    } catch (error) {
+      const errorMessage = error instanceof ApiError
+        ? `${error.message} (${error.status})`
+        : `[useFavorites] Retrieving favorites failed for ${userId}`;
+      setError(errorMessage);
+    } finally {
+      setLoading(false);
+    } return [];
```
chuckmilton: nit: This `return []` is the catch branch's implicit return path, in other words, the catch block doesn't return, so it falls through past `finally` to here. This works, but having it on the same line as the closing brace of `finally` can make it look like dead code. We should move it into the catch bock for clarity:
ex:
```typescript
    } catch (error) {
      const errorMessage = ...;
      setError(errorMessage);
      return [];       // <-- explicit here
    } finally {
      setLoading(false);
    }
```

zachfpadilla:
Thank you very much for the suggestions @chuckmilton—they are all excellent catches, especially in regards to the stub-like test cases in `useFavorites.test.ts`. 

I have since made the appropriate fixes, as well as adding an extra optimization in `MapScreen.tsx` & adding logout cleanup in `useFavorites.ts`. In updating the test cases, this also means that `@testing-library/react-native` has been added as a dev dependency.

In regards to the navigation flow, I implemented a navigation modal for a few reasons:
1. When selecting the favorites option, it grants users centralized access to all the lots they have favorited.
2. When selecting the recommended option, it is likely that there will be more than one promising lot at any given time (though it is true that there may be one that is most optimal), or even simply one/a few that the user would have a preference for if it's recommendable.
3. It allows the user to see the estimated occupancies and confidence scores of any favorited/recommended lots before one is selected.

That all being said, I would me more than fine with removing it & having the navigation button automatically select a lot on the user's behalf. It's completely my bad if this is the original & persisting intention.

Regardless, thank you again for your thorough review & I'm looking forward to receiving an even more thorough one from @DottL—character-by-character if possible!

chuckmilton:
> Thank you very much for the suggestions @chuckmilton—they are all excellent catches, especially in regards to the stub-like test cases in `useFavorites.test.ts`.
> 
> I have since made the appropriate fixes, as well as adding an extra optimization in `MapScreen.tsx` & adding logout cleanup in `useFavorites.ts`. In updating the test cases, this also means that `@testing-library/react-native` has been added as a dev dependency.
> 
> In regards to the navigation flow, I implemented a navigation modal for a few reasons:
> 
> 1. When selecting the favorites option, it grants users centralized access to all the lots they have favorited.
> 2. When selecting the recommended option, it is likely that there will be more than one promising lot at any given time (though it is true that there may be one that is most optimal), or even simply one/a few that the user would have a preference for if it's recommendable.
> 3. It allows the user to see the estimated occupancies and confidence scores of any favorited/recommended lots before one is selected.
> 
> That all being said, I would me more than fine with removing it & having the navigation button automatically select a lot on the user's behalf. It's completely my bad if this is the original & persisting intention.
> 
> Regardless, thank you again for your thorough review & I'm looking forward to receiving an even more thorough one from @DottL—character-by-character if possible!

Thank you for making these changes so promptly @zachfpadilla! I also appreciate your reasoning behind the navigation flow and because of that, I think it's a smart move to include it. 💯

PR #30, opened 2026-02-24
feat: Add parking validation system

lawrenceDegoma:
# Add Parking Validation System

## Overview
This PR introduces a backend system for validating parking behavior through mobile device sensor data. The system uses algorithmic analysis to determine whether users actually parked, drove through, or were searching for parking spots.

### Database Schema
- **ValidationEvent** model - Stores individual sensor readings with confidence scores
- **ParkingSession** model - Tracks complete parking experiences from entry to exit
- Database migration for new tables and relationships

### REST API Endpoints
- `POST /validation/events` - Record sensor data from mobile devices
- `POST /validation/sessions/start` - Begin parking sessions on lot entry
- `POST /validation/sessions/end` - End sessions with behavioral analysis
- `GET /validation/lots/:lotId/stats` - Real-time lot analytics and statistics

### Analysis
- **Speed Transition Analysis** - Detects parking vs driving patterns
- **Movement Pattern Recognition** - Identifies walking, stationary, driving behaviors
- **Bluetooth State Tracking** - Uses car connectivity for validation
- **Dwell Time Scoring** - Analyzes time spent in parking areas
- **Multi-factor Confidence Scoring** - Combines all signals for final determination

### Privacy & Security
- Device hash generation for user privacy
- No personal data stored, only behavioral patterns
- Configurable confidence thresholds

### Testing
- **29/29 tests passing** (100% coverage)
- **Controller tests** - Full REST API endpoint validation
- **Service tests** - Business logic and algorithm verification
- **Mock strategies** - Isolated unit testing with Prisma mocks

### Key Algorithms
- Confidence scoring based on event type, speed, GPS accuracy, Bluetooth state
- Session analysis combining speed transitions, movement patterns, and dwell time
- Real-time preliminary analysis during active sessions
- Final comprehensive analysis on session completion

### Validation Status Types
- `ANALYZING` - Session in progress
- `PARKED` - User successfully parked
- `DROVE_THROUGH` - User drove through without parking
- `SEARCHING` - User was looking for parking
- `UNKNOWN` - Insufficient data for determination

## Next Steps
This backend system is ready for mobile app integration. The next phase will involve:
- React Native sensor data collection
- Real-time geofence monitoring  
- Background location tracking
- Live occupancy updates

## Testing
```bash
pnpm test parking-validation  # Run all validation tests

### CODE REVIEW

File: apps/backend/src/parking-validation/dto/create-validation-event.dto.ts
Comment (chuckmilton) [inline apps/backend/src/parking-validation/dto/create-validation-event.dto.ts:1]: The global `ValidationPipe` in `main.ts` does nothing for `interface` DTOs. The rest of the codebase uses `class` + `class-validator` decorators (you can view an example in `create-occupancy-event.dto.ts`). Right now, any payload hits the service unchecked. We should fix it in this file,  [end-parking-session.dto.ts](https://github.com/chuckmilton/SharkPark/pull/30/changes#diff-af16e1957860a05e36fffb652bd7edd9dbf244c444b1c8ab15f7ef67d48bfe52), and [start-parking-session.dto.ts‎](https://github.com/chuckmilton/SharkPark/pull/30/changes#diff-99080164b0490668c696b77b15b935fbda7deb8378c34ff4792debd3e7f44460).

File: apps/backend/src/parking-validation/parking-validation.service.ts [hunk 1]
```diff
@@ -0,0 +1,378 @@
...
+      average_confidence: Math.round(averageConfidence * 100) / 100,
+    };
+  }
+
+  /**
+   * Generate privacy-preserving device hash
+   */
+  private generateDeviceHash(userId: string): string {
+    return crypto.createHash('sha256')
+      .update(userId + process.env.DEVICE_HASH_SALT || 'default_salt')
```
chuckmilton: This hash function will produce an operator precedence bug. It should be written like this: 
```typescript 
userId + (process.env.DEVICE_HASH_SALT || 'default_salt')
```

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+  constructor(private readonly prisma: PrismaService) {}
+
+  /**
+   * Record a validation event from mobile device sensors
+   * Includes algorithmic confidence scoring based on sensor data
+   */
+  async recordValidationEvent(dto: CreateValidationEventDto): Promise<ValidationEvent> {
+    try {
+      // Generate a device hash if not provided (for privacy)
+      const deviceHash = dto.deviceHash || this.generateDeviceHash(dto.userId);
```
chuckmilton: Test data shows `userId: 'test@csulb.edu'`. The existing module in `privacy.util.ts` takes a random device UUID, not an email. We should reuse that utility.

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+  /**
+   * Generate privacy-preserving device hash
+   */
+  private generateDeviceHash(userId: string): string {
+    return crypto.createHash('sha256')
+      .update(userId + process.env.DEVICE_HASH_SALT || 'default_salt')
+      .digest('hex')
+      .substring(0, 16);
```
chuckmilton: Reduces SHA-256 from 256 bits to 64 bits. The existing `privacy.util.ts` returns the full 64-char hex digest.

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+      const deviceHash = dto.deviceHash || this.generateDeviceHash(dto.userId);
+
+      // Look up the lot UUID from the lot_id string
+      const lot = await this.prisma.lot.findFirst({
+        where: { lot_id: dto.lotId },
+        select: { id: true },
+      });
+
+      if (!lot) {
+        throw new Error(`Lot ${dto.lotId} not found`);
```
chuckmilton: Generic Error instead of NestJS HTTP exceptions → clients get 500s:

`throw new Error('Lot ${dto.lotId} not found')` — should be `NotFoundException`

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+      const deviceHash = this.generateDeviceHash(dto.userId);
+      
+      // Look up the lot UUID from the lot_id string
+      const lot = await this.prisma.lot.findFirst({
+        where: { lot_id: dto.lotId },
+        select: { id: true },
+      });
+
+      if (!lot) {
+        throw new Error(`Lot ${dto.lotId} not found`);
```
chuckmilton: Generic Error instead of NestJS HTTP exceptions → clients get 500s:

`throw new Error('Lot ${dto.lotId} not found')` — should be `NotFoundException`

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+      // Find the active session
+      const session = await this.prisma.parkingSession.findFirst({
+        where: {
+          id: dto.sessionId,
+          exit_time: null,
+        },
+      });
+
+      if (!session) {
+        throw new Error('No active parking session found');
```
chuckmilton: Generic Error instead of NestJS HTTP exceptions → clients get 500s:

throw new Error('No active parking session found') — should be NotFoundException

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+    const since = new Date(Date.now() - hours * 60 * 60 * 1000);
+
+    // Look up the lot UUID from the lot_id string
+    const lot = await this.prisma.lot.findFirst({
+      where: { lot_id: lotId },
+      select: { id: true },
+    });
+
+    if (!lot) {
+      throw new Error(`Lot ${lotId} not found`);
```
chuckmilton: Generic Error instead of NestJS HTTP exceptions → clients get 500s:

throw new Error('Lot ${lotId} not found') — should be NotFoundException

Refer to `users.service.ts`

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+      }
+
+      // Perform comprehensive analysis
+      const analysisResult = await this.performSessionAnalysis(session.id);
+
+      // Update session with exit time and analysis results
+      await this.prisma.parkingSession.update({
+        where: { id: session.id },
+        data: {
+          exit_time: new Date(),
```
chuckmilton: `endParkingSession` ignores `dto.timestamp`

This always uses server time. Compare `parking-validation.service.ts` in `startParkingSession` which correctly does `dto.timestamp ? new Date(dto.timestamp) : new Date()`. Same pattern should be used for `exit_time`.

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+      throw error;
+    }
+  }
+
+  /**
+   * Start a new parking session when entering lot geofence
+   */
+  async startParkingSession(dto: StartParkingSessionDto): Promise<ParkingSession> {
+    try {
+      const deviceHash = this.generateDeviceHash(dto.userId);
```
chuckmilton: There's a device hash inconsistency between events and sessions. 
`const deviceHash = dto.deviceHash || this.generateDeviceHash(dto.userId)` allows custom hash.
`const deviceHash = this.generateDeviceHash(dto.userId)` ignores `dto.deviceHash`.

If a client sends a custom `deviceHash` for events, the session (created with a different hash) won't be found by `parking-validation.service.ts` which joins on `device_hash`. Analysis can silently fail.

apps/backend/src/parking-validation/parking-validation.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Get validation statistics for a specific lot
+   */
+  @Get('lots/:lotId/stats')
+  async getLotStats(
+    @Param('lotId') lotId: string,
+    @Query('hours') hours?: string,
+  ) {
+    const hoursNumber = hours ? parseInt(hours, 10) : 24;
```
chuckmilton: `parseInt('invalid', 10)` returns `NaN`, which then gives us `new Date(NaN)` in the service at `parking-validation.service.ts`. We should use `parseInt(hours, 10) || 24` or `ParseIntPipe`.

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+        data: {
+          device_hash: deviceHash,
+          lot_id: lot.id, // Use the UUID, not the lot_id string
+          event_type: dto.eventType,
+          timestamp: new Date(),
+          speed_mph: dto.speed,
+          accuracy_meters: dto.accuracy,
+          confidence_score: this.calculateConfidenceScore(dto),
+          bluetooth_state: dto.bluetoothState as BluetoothState,
+          raw_data: dto.rawData ? JSON.stringify(dto.rawData) : undefined,
```
chuckmilton: `raw_data` is double-serialized. It is `Json?` in `schema.prisma`. Prisma auto-serializes objects to JSONB. `JSON.stringify()` stores a JSON string inside JSONB (which makes it double-encoded)

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+        },
+      });
+
+      // Trigger analysis if there's an active session
+      await this.analyzeActiveSession(deviceHash, lot.id);
+
+      return event;
+    } catch (error) {
+      console.error('Error recording validation event:', error);
```
chuckmilton: We should use NestJS `Logger` instead of `console.error`

Every other service uses `private readonly logger = new Logger(...)`. You can find these in `occupancy-events.service.ts`.

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+          device_hash: deviceHash,
+          lot_id: lot.id, // Use the UUID, not the lot_id string
+          enter_time: dto.timestamp ? new Date(dto.timestamp) : new Date(),
+          validation_status: 'ANALYZING' as ValidationStatus,
+        },
+      });
+
+      return session;
+    } catch (error) {
+      console.error('Error starting parking session:', error);
```
chuckmilton: We should use NestJS `Logger` instead of `console.error`

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+          occupancy_contribution: analysisResult.contributesToOccupancy,
+          speed_transition_score: analysisResult.speedTransitionScore,
+          dwell_time_score: analysisResult.dwellTimeScore,
+          movement_pattern_score: analysisResult.movementPatternScore,
+          bluetooth_score: analysisResult.bluetoothScore,
+          validation_metadata: analysisResult.metadata,
+        },
+      });
+    } catch (error) {
+      console.error('Error ending parking session:', error);
```
chuckmilton: We should use NestJS `Logger` instead of `console.error`

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+      await this.prisma.parkingSession.update({
+        where: { id: activeSession.id },
+        data: {
+          validation_status: analysis.preliminaryStatus as ValidationStatus,
+          confidence_score: analysis.confidence,
+          validation_metadata: analysis.metadata,
+        },
+      });
+    } catch (error) {
+      console.error('Error analyzing active session:', error);
```
chuckmilton: We should use NestJS `Logger` instead of `console.error`

apps/backend/src/parking-validation/parking-validation.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+import { EndParkingSessionDto } from './dto/end-parking-session.dto';
+
+@Controller('validation')
+export class ParkingValidationController {
+  constructor(private readonly validationService: ParkingValidationService) {}
+
+  /**
+   * Record a new validation event from mobile device
+   */
+  @Post('events')
```
chuckmilton: No `@HttpCode` on POST endpoints. Compare `occupancy-events.controller.ts` which uses explicit `@HttpCode(HttpStatus.CREATED)`

This defaults to 201, which is fine

apps/backend/src/parking-validation/parking-validation.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+    return {
+      success: true,
+      data: event,
+    };
+  }
+
+  /**
+   * Start a parking session (called when entering lot geofence)
+   */
+  @Post('sessions/start')
```
chuckmilton: No `@HttpCode` on POST endpoints. Compare `occupancy-events.controller.ts` which uses explicit `@HttpCode(HttpStatus.CREATED)`

This defaults to 201, which is fine

apps/backend/src/parking-validation/parking-validation.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+    return {
+      success: true,
+      data: session,
+    };
+  }
+
+  /**
+   * End a parking session (called when exiting lot geofence)
+   */
+  @Post('sessions/end')
```
chuckmilton: No `@HttpCode` on POST endpoints. Compare `occupancy-events.controller.ts` which uses explicit `@HttpCode(HttpStatus.CREATED)`

This defaults to 201,  but this is an update, so we should make it 200

apps/backend/src/parking-validation/parking-validation.controller.ts
```diff
@@ -0,0 +1,62 @@
+import { Controller, Post, Get, Body, Param, Query } from '@nestjs/common';
+import { ParkingValidationService } from './parking-validation.service';
+import { CreateValidationEventDto } from './dto/create-validation-event.dto';
+import { StartParkingSessionDto } from './dto/start-parking-session.dto';
+import { EndParkingSessionDto } from './dto/end-parking-session.dto';
+
+@Controller('validation')
```
chuckmilton: nit: Route is `'validation'` instead of `'parking-validation'`. It would be best practice to keep it consistent. Every other module matches its domain

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+          movement_pattern_score: analysisResult.movementPatternScore,
+          bluetooth_score: analysisResult.bluetoothScore,
+          validation_metadata: analysisResult.metadata,
+        },
+      });
+    } catch (error) {
+      console.error('Error ending parking session:', error);
+      throw error;
+    }
+  }
```
chuckmilton: There's no transaction on `endParkingSession`. This reads session, performs analysis, updates session. 4 queries, no `$transaction`. Compare `occupancy-events.service.ts` which wraps the same pattern in `prisma.$transaction`.

apps/backend/src/parking-validation/parking-validation.service.ts
chuckmilton: There's no session dedup in `startParkingSession`. This creates a new session without checking if one already exists for this `device_hash + lot_id`. GPS bounce can fire duplicate ENTER events, which gives us duplicate sessions.

apps/backend/src/parking-validation/dto/create-validation-event.dto.ts
```diff
@@ -0,0 +1,15 @@
+import { ValidationEventType, BluetoothState } from '@prisma/client';
+
+export interface CreateValidationEventDto {
+  userId: string;
+  lotId: string;
+  eventType: ValidationEventType;
+  latitude: number;
+  longitude: number;
```
chuckmilton: `latitude` and `longitude` are accepted but never used in here. Is there a reason why? If we're not planning to have this service reference these fields, we should either remove or document.

apps/backend/src/parking-validation/dto/start-parking-session.dto.ts
```diff
@@ -0,0 +1,8 @@
+export interface StartParkingSessionDto {
+  userId: string;
+  lotId: string;
+  latitude: number;
+  longitude: number;
```
chuckmilton: `latitude` and `longitude` are accepted but never used in here. Is there a reason why? If we're not planning to have this service reference these fields, we should either remove or document.

apps/backend/src/parking-validation/dto/end-parking-session.dto.ts
```diff
@@ -0,0 +1,6 @@
+export interface EndParkingSessionDto {
+  sessionId: string;
+  latitude?: number;
+  longitude?: number;
```
chuckmilton: `latitude` and `longitude` are accepted but never used in here. Is there a reason why? If we're not planning to have this service reference these fields, we should either remove or document.

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+        where: { id: activeSession.id },
+        data: {
+          validation_status: analysis.preliminaryStatus as ValidationStatus,
+          confidence_score: analysis.confidence,
+          validation_metadata: analysis.metadata,
+        },
+      });
+    } catch (error) {
+      console.error('Error analyzing active session:', error);
+    }
```
chuckmilton: nit: This can swallow errors silently. `parking-validation.service.ts` catches, logs, and returns. Caller at `parking-validation.service.ts` never knows analysis failed.

apps/backend/test/parking-validation.service.spec.ts
```diff
@@ -0,0 +1,486 @@
...
+
+    it('should calculate appropriate confidence scores for different event types', async () => {
+      const testCases = [
+        { eventType: 'SPEED_CHANGE', speed: 0.5, expectedMinConfidence: 0.8 },
+        { eventType: 'STATIONARY', speed: 0, expectedMinConfidence: 0.85 },
+        { eventType: 'WALKING', speed: 3, expectedMinConfidence: 0.5 },
+        { eventType: 'DRIVING', speed: 15, expectedMinConfidence: 0.7 },
+        { eventType: 'BLUETOOTH_CONNECT', expectedMinConfidence: 0.6 },
+        { eventType: 'GEOFENCE_ENTER', expectedMinConfidence: 0.7 },
+      ];
```
chuckmilton: nit: declares `expectedMinConfidence: 0.7` for `DRIVING` but the actual score would be `~0.4`. The assertion only checks `score >= 0 && score <= 1`.

apps/backend/src/parking-validation/parking-validation.service.ts
```diff
@@ -0,0 +1,378 @@
...
+  /**
+   * Analyze active parking session based on accumulated validation events
+   */
+  private async analyzeActiveSession(deviceHash: string, lotUuid: string): Promise<void> {
+    try {
+      const activeSession = await this.prisma.parkingSession.findFirst({
+        where: {
+          device_hash: deviceHash,
+          lot_id: lotUuid, // This is already the UUID
+          exit_time: null,
```
chuckmilton: This queries `{ device_hash, lot_id, exit_time: null }` but I believe the schema only has indexes on `(device_hash, enter_time)` and `(lot_id, enter_time)`. Checkout `schema.prisma`.

apps/mobile/ios/Podfile.lock
chuckmilton, line 1: We should revert this file. The diff shows CocoaPods downgraded from 1.16.2 to 1.15.2 and a lot of pod hashes changed.

chuckmilton (changes requested):
Very solid PR @lawrenceDegoma, I'm impressed with your work! Filtering drive-throughs from real parks would definitely improve occupancy accuracy. However I do have some concerns about this approach.

All the sensor data being analyzed (speed, GPS accuracy, Bluetooth state, motion patterns) originates on the phone. The `calculateConfidenceScore` and `analyzeEventPatterns` functions don't reference any server-side data, they only use what came from the client. Feel free to correct me if I'm wrong, but instead of phone → network → server stores → server analyzes → server stores result, the phone could just run the same analysis locally and send a single classification with the ENTER/EXIT event we already send.

Privacy is a big concern for our app. Our entire codebase is built around never storing user location data. The `locationService` header says "never stores actual user coordinates," the `GeofenceEvent` type has "NO location coordinates - privacy first," we SHA-256 hash device IDs before storage, and the infrastructure README explicitly calls out "no PII in events." This PR stores userId, lat/lng, speed timelines, Bluetooth state sequences, and movement patterns per user on our server. Because that's a behavioral tracking profile, it goes against what the rest of the app was designed to avoid.

Some other things:
- occupancy_contribution gets computed but never feeds back into current_occupancy or the snapshot pipeline, so the validation system and occupancy system run in parallel without connecting
- The ML pipeline trains on occupancy_snapshots joined with calendar/weather data. The new validation tables aren't referenced anywhere in model design or feature engineering
- Streaming multiple sensor events per user per session could be 10-50x our current write volume (~4 writes/sec at peak), which changes our Aurora scaling and cost assumptions

I think the better approach is running the behavioral analysis on the client and adding a validation_status and confidence_score to the existing OccupancyEvent. The phone classifies locally, sends one enriched event, and we use the status to decide whether to count it toward occupancy. No new tables, no raw sensor data over the wire, privacy model stays intact.

If I'm misunderstanding something here, feel free to let me know. I'd be more than happy to discuss this further 🙏

PR #31, opened 2026-02-25
ML Core Modules

DottL:
# ML Core Modules — Data, Features, Model, Evaluation

## Overview

Core ML modules for short-term parking occupancy prediction at CSULB. This PR adds the foundational building blocks — synthetic data generation, feature engineering, XGBoost model with quantile regression, naive baselines, evaluation metrics, and MLflow experiment tracking.

These modules are consumed by the CLI scripts (train, evaluate, promote, predict) which will be added in a follow-up PR along with database write utilities and workflow tests.

### Data Generation
- Synthetic data generator pulling lot metadata (IDs, capacities, types) from PostgreSQL
- ~237K occupancy snapshots (~8,450 per lot × 28 lots) for Fall 2025 semester — one per 15-min slot across the semester
- Heuristic academic calendar with rule-based semester/break detection for CSULB
- `is_cold_start` flag to distinguish synthetic from real data during blended transition

### Feature Engineering
- Shared base utilities: cyclical hour/day encodings, timestamp normalization, validation
- Short-term features: lag features at t-15min, t-30min, t-45min, t-60min
- Momentum (rate of change between current and lag_1)
- Academic context: `academic_period`, `week_of_semester`, `is_campus_open`
- Multi-horizon support via `hours_ahead` feature (single model handles 1–14 hour predictions)
- Separate training and inference feature builders

### Model
- XGBoost regression model with save/load support
- Quantile regression (10th/90th percentile) for data-driven confidence intervals
- Naive baselines for comparison: persistence and historical average
- Single model architecture handles all prediction horizons

### Evaluation
- Metrics: MAE, RMSE, MAPE
- Comparison framework: candidate model vs baselines vs current production model
- Promotion criteria: must beat all baselines on MAE, reduce MAE ≥5% vs production or improve directional accuracy ≥ 3 percentage points

### MLflow Integration
- Experiment tracking with logged hyperparameters and metrics per run
- Model registry with `short-term-production` alias
- Local-only (`mlruns/` directory) — no remote server required

### Exploratory Notebook
- `notebooks/check_synthetic.ipynb` — sanity-check visualizations for synthetic data distribution

### Source
- `src/academic_calendar.py` - Rule-based academic calendar for CSULB 
- `src/config.py` - Operating hours (7–21), snapshot interval (15 min), database URL 
- `src/data/synthetic.py` - Synthetic occupancy data generator 
- `src/features/__init__.py` - Feature engineering package init with public API exports 
- `src/features/base.py` - Shared utilities: time encoding, validation, normalization 
- `src/features/short_term.py` - Lag features, momentum, training + inference feature builders 
- `src/models/short_term.py` - XGBoost regression model class with quantile regression 
- `src/models/baselines.py` - Naive baselines (persistence, historical average) 
- `src/evaluation/metrics.py` - MAE, RMSE, MAPE metric functions 

### Docs
- `README.md` - Full ML service documentation (setup, workflow, data flow, deployment)

### Notebooks
- `notebooks/check_synthetic.ipynb` - Sanity-check visualizations for synthetic data distribution

### Config
- `pyproject.toml` - Python project config (dependencies, dev tools)
- `.gitignore` - ML-specific ignores (mlruns/, data/*.parquet, etc.)
- `data/.gitkeep` - Keep data directory in git

### Tests
- `tests/data/test_synthetic.py` - Synthetic data generation tests
- `tests/features/test_academic_calendar.py` - Academic calendar logic tests
- `tests/features/test_base.py` - Base feature utility tests
- `tests/models/test_baselines.py` - Baseline model tests
- `tests/models/test_short_term.py` - XGBoost model training/prediction tests
- `tests/evaluation/test_metrics.py` - Evaluation metrics (MAE, RMSE, MAPE) tests

## Notes

- **Predicted occupancy is an integer count**, not a normalized rate. The model predicts a rate (0.0–1.0) internally, then multiplies by lot capacity to produce a count.
- **Schema change**: Synthetic data columns changed from `source` (always "synthetic") to `is_cold_start` (boolean) + `academic_period`, `week_of_semester`, `is_campus_open`.
- Models stored locally in `mlruns/` during development. S3 only needed when deploying to Lambda.
- **No CLI scripts in this PR** — train, evaluate, promote, predict scripts will be in a follow-up PR along with database write utilities, model comparison framework (`evaluation/compare.py`).

## What's Next
In progress:
- CLI scripts (train, evaluate, promote, predict)
- Database utilities for reading real data from PostgreSQL
- Model comparison framework (candidate vs baselines vs production)
- End-to-end workflow and integration tests


services/ml/src/academic_calendar.py
```diff
@@ -0,0 +1,605 @@
...
+    period_map = {
+        "classes": "regular",
+        "finals": "finals",
+        "break": "break",
+        "reading_day": "regular",
+        "pre_classes": "regular",
+        "intersession": "intersession",
+        "between_semesters": "off_session",
+    }
+    return period_map.get(period_type, "regular")
```
chuckmilton: This function returns 5 distinct values (`regular`, `finals`, `break`, `intersession`, `off_session`), but the design doc and infrastructure DB CHECK constraint both specify 4 (`regular`, `finals`, `break`, `summer`). The code never returns "summer", intersession/between-semester periods map to `"intersession"` and `"off_session"` instead.

This means two things:
1. writing these snapshots to our SQL database (Aurora) would violate the CHECK constraint
2. the model trains on categories that diverge from the design spec

Also the Prisma schema comment on `OccupancySnapshot.academic_period` says `"FALL", "SPRING", "SUMMER", "BREAK"` which might be misleading. Those are `AcademicCalendar.period_type` values (semester labels), not academic period values. It should be updated to `"regular", "finals", "break", "summer"` to match the design doc.

We should either update the `period_map` to map `intersession -> summer` and `between_semesters -> summer` (which matches the design doc), or we update the design doc and DB constraint to include new categories.

services/ml/src/features/base.py
```diff
@@ -0,0 +1,302 @@
...
+        raise ValueError(f"Missing required columns: {missing}")
+
+    df = df.copy()
+
+    # Remove rows with null values in required columns
+    df = df.dropna(subset=required_cols)
+
+    # Filter out low-confidence snapshots
+    if "confidence" in df.columns:
+        df = df[df["confidence"].isin(["HIGH", "MEDIUM"])]
```
chuckmilton: During cold-start, most real snapshots will have `confidence = "LOW"` because there won't be enough devices reporting. This filter will silently discard them, which potentially leaves too few rows for training or inference.

What we could do is make the confidence threshold configurable (e.g., a `min_confidence` parameter defaulting to `["HIGH", "MEDIUM"]`). for training you probably want HIGH/MEDIUM only, but inference may need to accept LOW. At minimum, we can log a warning when >50% of rows are dropped so this doesn't become a silent data leak.

services/ml/src/data/synthetic.py
```diff
@@ -0,0 +1,793 @@
...
+    SNAPSHOT_INTERVAL_MINUTES,
+)
+
+# ---------------------------------------------------------------------------
+# Derive calendar dates from academic_calendar (Fall 2025)
+# ---------------------------------------------------------------------------
+_fall = ACADEMIC_CALENDARS["2025-2026"]["fall"]
+
+SEMESTER_START = datetime.combine(_fall["semester_start"], datetime.min.time())
+SEMESTER_END = datetime.combine(_fall["semester_end"], datetime.min.time())
```
chuckmilton: These module-level constants are evaluated at import time, so all of `synthetic.py` is locked to Fall 2025. To generate data for Spring 2026 or any other semester, you'd have to edit source code.

We should move semester selection into the CLI (e.g., `--semester fall-2025` / `--semester spring-2026`) and pass the resolved calendar dict into generate_all_data() as a parameter. The module-level constants can become function-local or fields on a config dataclass.

services/ml/src/data/synthetic.py
```diff
@@ -0,0 +1,793 @@
...
+    Returns:
+        Dict matching OccupancySnapshot schema
+    """
+    # Compute calendar features
+    date_only = timestamp.replace(hour=0, minute=0, second=0, microsecond=0)
+    campus_open = is_campus_open(timestamp)
+    academic_period = get_academic_period(timestamp)
+    week_of_sem, _, _ = get_week_of_semester(timestamp.date())
+
+    if date_only in CAMPUS_CLOSURES:
```
chuckmilton: `date_only` is a `datetime` zeroed to midnight and `CAMPUS_CLOSURES` is a `list[datetime]`, so this works today. But `is_campus_open()` internally converts to `.date()` and compares against `date` objects. Mixing `datetime` and `date` across the codebase is fragile because if one side changes representation, the comparison silently fails. Also, `in` on a list is O(n) per snapshot.

We should use `timestamp.date()` consistently, make `CAMPUS_CLOSURES` a `set[date]` (O(1) lookup), and do the same for `NO_CLASSES_CAMPUS_OPEN`.

services/ml/src/data/synthetic.py
```diff
@@ -0,0 +1,793 @@
...
+        - 4-6pm: 0.85 -> 0.25 (main departure window)
+        - 6-9pm: 0.25 -> 0.05 (stragglers, evening staff)
+
+    Returns base occupancy rate (0.0-1.0) before noise/multipliers.
+    """
+    time_decimal = hour + minute / 60.0
+
+    # Buffer hours
+    if hour < OPERATING_START_HOUR or hour >= OPERATING_END_HOUR:
+        return 0.0
```
chuckmilton: `OPERATING_END_HOUR` is 21. The guard `hour >= 21` fires before the evening interpolation `[9, 21]` can reach its endpoint. So at exactly 21:00 occupancy jumps from ~0.08 (at 20:45) to 0.0. Since `PREDICTION_HOURS` includes 21, the model will train on hour-21 snapshots that are always ~0.0.

Functionally, this is most likely fine, but I just wanted to note that the interpolation's endpoint is unreachable. If that's intentional, a brief comment would help future readers. If 21:00 should produce ~0.05, change the guard to `hour > OPERATING_END_HOUR`.

services/ml/src/models/short_term.py
```diff
@@ -0,0 +1,400 @@
...
+        with mlflow.start_run(run_name="short-term-training") as run:
+            if params:
+                mlflow.log_params(params)
+            mlflow.log_metrics(metrics)
+
+            # Save model artifact
+            from urllib.parse import urlparse, unquote
+
+            parsed = urlparse(run.info.artifact_uri)
+            artifact_path = Path(unquote(parsed.path)).resolve()
```
chuckmilton: This works for default local `mlruns/` backend, but `artifact_uri` can be `s3://...`, `gs://...`, or `dbfs://...` with remote MLflow setups. `Path(unquote(parsed.path))` will produce garbage paths for non-file URIs.

We should use `mlflow.artifacts.download.artifacts()` for loading or `mlflow.log_artifact()` for saving instead of manual path parsing. If this is intentionally local-only fow now, add a guard: `assert parsed.scheme in ("", "file"), f"Remote artifact URIs not supported: {run.info.artifact_uri}"` so it fails loud instead of silently corrupting paths.

services/ml/src/features/short_term.py
```diff
@@ -0,0 +1,313 @@
...
+                    "occupancy_rate_lag_1": row["occupancy_rate_lag_1"],
+                    "occupancy_rate_lag_2": row["occupancy_rate_lag_2"],
+                    "occupancy_rate_lag_3": row["occupancy_rate_lag_3"],
+                    "occupancy_rate_lag_4": row["occupancy_rate_lag_4"],
+                    "momentum": row["momentum"],
+                    "target_hour": target_hour,
+                    "hours_ahead": hours_ahead,
+                    "target_occupancy_rate": target_val,
+                }
+            )
```
chuckmilton: For the Fall semester synthetic dataset (~137K snapshots x ~7 avg remaining hours), this generated ~1.6 million rows via `iterrows()` + dict appends. `iterrows()` is extremely slow in pandas because it boxes each row into a Series object.

We should vectorize with a cross-join instead:
```python
hours_df = pd.DataFrame({"target_hour": PREDICTION_HOURS})
expanded = df.merge(hours_df, how="cross")
expanded = expanded[expanded["target_hour"] > expanded["hour"]]
expanded["hours_ahead"] = expanded["target_hour"] - expanded["hour"]
# Merge actuals
actuals_df = df[["lot_id", "date", "hour", "occupancy_rate"]].rename(
    columns={"hour": "target_hour", "occupancy_rate": "target_occupancy_rate"}
)
expanded = expanded.merge(actuals_df, on=["lot_id", "date", "target_hour"], how="inner")
```

Not blocking, but will matter as dataset size grows.

services/ml/src/models/baselines.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Note: predict() is scaffolding for future API inference use.
+    Currently only evaluate() is wired up (via compare.py).
+    """
+
+    def predict(self, *args, **kwargs):
+        # TODO: Wire up for real-time API inference.
+        # Should take recent snapshots + lot_ids, return the latest
+        # occupancy_rate repeated for each remaining PREDICTION_HOURS.
+        raise NotImplementedError("predict() not yet wired up for inference")
```
chuckmilton: Both `PersistenceBaseline` and `MajorityClassBaseline` have this. The docstrings explain it's scaffolding and `evaluate()` is the only wired path, which is fine for now. However, anyone calling `.predict()` from a future API handler will get an unhandled exception with no context.

This is a low priority, but we could consider either implementing the trivial versions now (persistence = return current rate; majority = return median) or tracking this as a follow-up issue so it doesn't get forgotten when the API inference path is built.

services/ml/tests/data/__init__.py
chuckmilton, line 1: Several test files independently create similar sample DataFrames (snapshot data with lot_id, timestamp, occupancy, etc.). There's no shared `conftest.py` with pytest fixtures.

We should add a `tests/conftests.py` with shared fixtures like `sample_snapshot_df`, `sample_training_features`, etc. This reduces duplication and makes it easier to update test data if the schema changes.

services/ml/tests/features/__init__.py
chuckmilton, line 1: Only `src/features/__init__.py` defines `__all__`. The other modules (`academic_calendar.py`, `config.py`, `data/synthetic.py`, `models/short_term.py`, `models/baselines.py`, `evalutation/metrics.py`) expose their full namespace on wildcard imports.

We should add `__all__` to each module listing the intended public API. Low priority, but good practice as the  codebase grows.

chuckmilton (commented):
Great work on the ML Core Modules! Overall, this is a really solid foundation for the ML pipeline. The code is well-organized, thoroughly documented, and the architecture clearly follows the design doc. The academic calendar module is impressively detaied, and the XGBoost model + baseline comparison setup exactly aligns with what the design doc calls for.

That being said, I left some suggestions for you to review and to possibly make revisions on. The first three issues are the ones I'd definitely want to try to fix before merge (the academic_period mismatch will break real data ingestion, the confidence filter will silently drop cold-start data, and the hardcoded semester makes the generator inflexible). Everything else is either a robustness improvement or a minor nit that can be a follow-up.

Amazing job @DottL!

DottL:
Feedback addressed 🎤. I've updated the schema per our discussion @chuckmilton. Happy reading!! 😄

-----
### Summary
- **Configurable confidence filtering** — allow `min_confidence` override for cold-start inference scenarios
- **Vectorized training features** — replace `iterrows` with cross-join for faster `prepare_training_features`
- **MLflow artifact portability** — use `log_artifacts`/`download_artifacts` instead of parsing artifact URIs directly
- **Semester-agnostic synthetic data** — add `--semester` CLI flag to the generator so it's not hardcoded to a single term
- **Split `academic_period`** — add new `semester` column (e.g. Spring 2025) alongside a more granular `academic_period` (e.g. finals, midterms), with a Prisma migration to match
- **Source tagging** — synthetic data rows are now tagged with `source = 'synthetic'` for traceability
- **Shared test fixtures** — added `conftest.py` with reusable fixtures reflecting the new semester/period schema

### Files changed
**Backend**
- `apps/backend/prisma/migrations/.../migration.sql` — add `semester` column, update training index
- `apps/backend/prisma/schema.prisma` — add `semester` field, refine `academic_period` to granular periods

**ML Service — Core**
- `services/ml/src/academic_calendar.py` — refactored calendar logic for semester awareness
- `services/ml/src/data/synthetic.py` — semester-agnostic generator with `--semester` CLI flag, source tagging
- `services/ml/src/features/base.py` — vectorized `prepare_training_features`
- `services/ml/src/features/short_term.py` — updated feature extraction for new period schema
- `services/ml/src/features/__init__.py` — removed stale exports
- `services/ml/src/models/short_term.py` — configurable `min_confidence`, MLflow artifact portability

**ML Service — Tests**
- `services/ml/tests/conftest.py` — updated shared fixtures for new schema
- `services/ml/tests/data/test_synthetic.py` — tests for semester flag & source tagging
- `services/ml/tests/features/test_academic_calendar.py` — updated calendar tests
- `services/ml/tests/features/test_base.py` — new tests for vectorized features
- `services/ml/tests/models/test_short_term.py` — updated model tests

**Docs**
- `services/ml/Model_Design.md` — updated design doc
- `services/ml/README.md` — updated usage instructions

chuckmilton:
Great work on those revisions! Everything looks good to me now 🤓👍

PR #60, opened 2026-03-12
Feature/Short-Term Forecast

DottL:
# Short-Term Forecast Pipeline

## Overview

Adds the CLI pipeline (train/evaluate/promote/predict), database read/write utilities, model comparison logic, sample weighting for cold-start transition, and cold-start confidence interval widening.

### Sample Weighting (Cold-Start Transition)
- `_build_sample_weights()` classifies rows into 3 tiers — **synthetic**, **real cold-start**, **real established** — and volume-normalizes so each group's influence matches its weight parameter regardless of row count
- Weights applied to all 3 XGBoost models (median, lower quantile, upper quantile)
- Off by default — weights are all 1.0 (uniform). Only activates when training data has a `_source` column and at least one weight is set to something other than 1.0 (e.g. `synthetic_weight=0.3`)

### Cold-Start Confidence Interval Widening
- Cold-start lots get 1.5x wider confidence intervals (`COLD_START_CI_MULTIPLIER`), spread symmetrically around median and clipped to [0, 1]
- Applied at inference time only — the model itself doesn't know about cold-start status

### Database Utilities (`src/data/db.py`)
- PostgreSQL connection helper, lot_id mapping (human-readable → CUID), snapshot reads, and batch prediction writes with delete-before-insert cleanup
- **Note:** `write_predictions()` deletes existing predictions for the same lots and inserts new ones in a single transaction — if the script crashes mid-write, the transaction rolls back and old predictions remain intact. If we implement drift detection later, we'd switch to append-with-TTL instead of delete-before-insert so we can retain old predictions and compare them against actuals.

### Training Data Blending (`scripts/train.py`)
- `--include-real` flag loads real data from PostgreSQL and blends it with synthetic data — real rows replace overlapping synthetic rows for the same dates, so the model trains on real data where available and falls back to synthetic elsewhere
- `--synthetic-weight` and `--cold-start-weight` control the relative influence of each data tier (feeds into `_build_sample_weights()` above)

### Fair Evaluation (`scripts/evaluate.py`)
- Both candidate and production models are evaluated on the **same test set** (derived from the candidate's training data artifact) for an apples-to-apples comparison
- Reproduces the same temporal split used during training, then re-scores the production model on that test set to avoid data drift bias

### Model Comparison & Promotion (`src/evaluation/compare.py`)
- Promotion criteria: must beat all baselines on MAE, then either reduce MAE ≥5% vs production or improve directional accuracy ≥3pp. First deployment auto-promotes if candidate beats baselines.
- **Coverage-based baseline gates** — which baselines are required depends on real data coverage:
  - < 30% → skip baseline comparison entirely
  - 30–60% → persistence baseline only
  - \> 60% (≥4 obs/combo) → all baselines
- **Why aggregate coverage (not per-lot)?** There's one global XGBoost, so promotion is "promote this model or don't." Per-lot coverage would only make sense with per-lot models.

### CLI Scripts
```
train.py  ->  evaluate.py  ->  promote.py  ->  predict.py
  (MLflow run ID passed between each step)
```
| Script | Command | What it does |
|--------|---------|--------------|
| Train | `python -m scripts.train` | Train XGBoost on synthetic (+ optional real) data, log to MLflow |
| Evaluate | `python -m scripts.evaluate --run-id <id>` | Compare candidate against baselines + production |
| Promote | `python -m scripts.promote --run-id <id>` | Register model as `short-term-production` in MLflow |
| Predict | `python -m scripts.predict` | Batch inference, writes to PostgreSQL |
| Predict (full day) | `python -m scripts.predict --start-of-day` | Predict all hours (7-21) regardless of current time. Use for scheduled/nightly runs |
| Predict (local) | `python -m scripts.predict --write-local` | Also write predictions to local CSV |
| Check DB | `python -m scripts.check_predictions` | Temporary debugging utility — inspect predictions in the database |

## Up Next
- **Long-term forecast model** — Two-stage hybrid (4-week rolling historical baseline + XGBoost deviation adjustment) for 7-day-ahead hourly predictions per lot
- **`HistoricalAverage` and `SameDayLastWeek` baselines** — Add to `compare.py` once long-term model exists
- **`--model-type` flag for CLI scripts** — Let train/evaluate/promote/predict select between short-term and long-term models
- **Hyperparameter CLI flags** — `DEFAULT_HYPERPARAMS` dict and `hyperparams` override in `ShortTermModel.train()` are already in place. Still need CLI flags in `train.py` (e.g. `--n-estimators 300 --learning-rate 0.05`) to expose them from the command line


services/ml/src/data/db.py
```diff
@@ -0,0 +1,272 @@
...
+        query = text("""
+            SELECT lot_id, timestamp, occupancy, available, occupancy_rate,
+                   confidence, is_cold_start, academic_period,
+                   week_of_semester, is_campus_open, semester
+            FROM occupancy_snapshots
+            WHERE timestamp >= NOW() - INTERVAL :hours
+            ORDER BY timestamp
+        """)
+
+        df = pd.read_sql(query, conn, params={"hours": f"{lookback_hours} hours"})
```
chuckmilton: `WHERE timestamp >= NOW() - INTERVAL :hours` with a bound string is fragile/invalid in Postgres SQL parsing for many drivers. Live prediction path can fail before feature generation when reading recent snapshots.

We should use `make_interval(hours => :lookback_hours)`, or `NOW() - (:lookback_hours || ' hours')::interval`, or arithmetic with `INTERVAL '1 hour' * :lookback_hours`.

services/ml/src/evaluation/compare.py
```diff
@@ -0,0 +1,360 @@
...
+    counts = df.groupby(["lot_id", "_dow", "_hour"]).size()
+
+    # Total possible combinations across ALL lots in the system
+    total_combos = total_lots * 7 * 15  # lots * 7 days × 15 prediction hours (7-21)
+
+    if total_combos == 0:
+        return 0.0
+
+    qualifying = (counts >= min_observations).sum()
+    return qualifying / total_combos
```
chuckmilton, line 108: Coverage gate logic in comparison can overcount and promote with inflated confidence. Numerator counts all observed hours, denominator assumes only 15 operating hours (7-21). Computed coverage can exceed intended scale and unlock stricter baseline gates too early, weakening promotion correctness.

We should filter to operating hours before grouping, or adjust denominator to the actual hour domain used by numerator

services/ml/scripts/evaluate.py
```diff
@@ -0,0 +1,182 @@
...
+        logger.info(
+            "Re-evaluating production model (run %s) on candidate test set...", run_id
+        )
+        prod_model = ShortTermModel.load_mlflow(run_id)
+        prod_preds = prod_model.predict(test_features)
+        actuals = test_features["target_occupancy_rate"].values
+        return compute_metrics(actuals, prod_preds)
+
+    except mlflow.exceptions.MlflowException:
+        logger.info("No production model registered — first deployment.")
```
chuckmilton: Evaluation can silently mask production-eval failures as “first deployment”. Broad except mlflow.exceptions.MlflowException treats any MLflow exception as no production model. Registry/network/auth outages can incorrectly bypass production comparison and bias toward promotion.

We should distinguish “alias/model not found” from transient/system errors; only return None for true not-found.

services/ml/scripts/train.py
```diff
@@ -0,0 +1,222 @@
...
+            df_synthetic["timestamp"] = pd.to_datetime(df_synthetic["timestamp"]).dt.tz_localize(None)
+
+            # Round to 15-min intervals to match on snapshot slots
+            df_real["_slot"] = df_real["timestamp"].dt.floor("15min")
+            df_synthetic["_slot"] = df_synthetic["timestamp"].dt.floor("15min")
+
+            real_keys = set(zip(df_real["lot_id"], df_real["_slot"]))
+            mask = ~pd.Series(
+                list(zip(df_synthetic["lot_id"], df_synthetic["_slot"]))
+            ).apply(lambda x: x in real_keys)
```
chuckmilton: Train synthetic/real overlap replacement is O(n) Python-heavy and not scalable. Building Python tuple sets + row-wise apply(lambda) for membership is expensive on large datasets, which means unnecessary CPU/memory cost during training as data grows.

We should replace with vectorized anti-join/merge on [lot_id, _slot] or MultiIndex difference.

services/ml/scripts/check_predictions.py
```diff
@@ -0,0 +1,50 @@
...
+            return
+
+        cur.execute("""
+            SELECT lot_id, predicted_at, target_time,
+                   predicted_occupancy, confidence_lower, confidence_upper,
+                   model_version
+            FROM predictions_short_term
+            ORDER BY lot_id, target_time            
+        """)
+        rows = cur.fetchall()
```
chuckmilton: Debug script can become a data-leak/perf hazard in shared logs. It fetches and prints all rows without LIMIT or pagination, which means terminal/log flooding, potential exposure of full prediction history in CI/shared environments, and avoidable memory pressure.

We should add --limit and optional filters (lot/date/version), default to small recent window.

services/ml/scripts/check_predictions.py
```diff
@@ -0,0 +1,50 @@
...
+
+Usage (from services/ml/):
+    python -m scripts.check_predictions
+"""
+
+from src.data.db import get_connection
+
+
+def main():
+    conn = get_connection()
```
chuckmilton: Robustness gap in DB debug helper connection lifecycle (line 12 and line 46). Manual close pattern instead of context manager. Exceptions before final close can leak connection.

Use with `get_connection()` as conn: and simplify the early-return branch.

services/ml/scripts/train.py
chuckmilton, line 1: In `train.py` and `short-term.py`, your PR states that sample weighting is off by default, but the CLI defaults set both of these weights as values that are not 1. This will activate once `_source` exists, which will change the training behavior unless they're set to 1.0. If this is intentional, then we can leave it as is since we don't have `_source` yet, but if not then we should set the defaults to 1.0 or  update the docs to reflect that weighting is on by default.

services/ml/scripts/promote.py
chuckmilton, line 1: nit: promote script behavior and docstring are slightly inconsistent. Doc says transitions to Production stage; implementation uses model alias only (modern MLflow approach). We should update wording to “sets production alias” everywhere.

services/ml/src/config.py
chuckmilton, line 1: `config.py` hardcodes a local credentialed `DATABASE_URL` fallback. Fine for dev, but in production it risks misconfig if env vars are missing. Prefer fail‑fast or require explicit env.

services/ml/tests/evaluation/test_compare.py
chuckmilton, line 1: `compare_models()` calls `get_total_lot_count()` unconditionally. That hits Postgres even when using `--data-path` (purely local evaluation), which makes evaluation non‑reproducible in offline or CI contexts

services/ml/scripts/predict.py
```diff
@@ -0,0 +1,245 @@
...
+    preds, preds_lower, preds_upper = model.predict_quantiles(features)
+
+    # Derive capacity from snapshots so inference stays self-contained
+    # (works with both --data-path parquet and live DB, no extra query needed).
+    # Falls back to 200 in _build_prediction_df if a lot_id is missing.
+    lot_capacities = (
+        df.groupby("lot_id")
+        .apply(
+            lambda g: (g["occupancy"] + g["available"]).median(), include_groups=False
+        )
```
chuckmilton: Small optimization in inference capacity calculation. `groupby.apply(lambda ...)` is slower than vectorized expression. This is modest but avoidable overhead. We should compute temp column `occupancy_plus_available` then `groupby.median()` directly.

services/ml/src/data/db.py
```diff
@@ -0,0 +1,272 @@
...
+
+        if df.empty:
+            raise RuntimeError(
+                f"No snapshots found in the last {lookback_hours} hours. "
+                "Is the backend scheduler running?\n"
+                "For local dev, use --data-path to load from a parquet file instead:\n"
+                "  python -m scripts.predict --data-path data/synthetic_fall-2025.parquet --start-of-day"
+            )
+
+        df["lot_id"] = df["lot_id"].map(reverse_map)
```
chuckmilton, line 122: Missing guard for unknown lot mapping on DB reads. After reverse-mapping lot IDs, unmapped IDs become null silently. This can propagate bad rows into feature generation and fail later in less obvious ways. We should validate for null `lot_id` immediately and raise a clear error listing unknown IDs.

chuckmilton (changes requested):
Great work on this PR overall. The pipeline shape is solid and the flow is clear: train → evaluate → promote → predict. I especially like the fair-eval approach (candidate and production scored on the same test set), the cold-start uncertainty handling, and the effort to add end-to-end tests.

I do think we should address a few issues before merge:

1. Blockers
- Evaluation currently catches broad MLflow exceptions and can treat real registry/network/auth failures as “no production model,” which could incorrectly bias promotion decisions.
- Coverage gating logic mixes numerator/denominator hour scopes, which can inflate coverage and unlock stricter baseline gates too early.
- Recent snapshot SQL interval handling looks fragile and may fail at runtime depending on Postgres/driver parsing.

2. Should-fix
- Debug prediction checker fetches/prints all rows with no limit/filter, which is risky for shared logs and can be heavy on memory.
- Debug checker uses manual connection close; moving to context-manager style improves safety on exceptions.
- Synthetic/real overlap replacement in training uses Python-level set + row-wise lambda checks; we should switch to a vectorized anti-join for scalability.
- Weighting defaults in train CLI do not match “off by default” behavior described in docs/design. Either set defaults to 1.0 or update docs explicitly.

3. Nice-to-have optimizations
- In prediction capacity derivation, replace groupby apply(lambda) with a fully vectorized approach.
- Align wording in promotion docs/comments: implementation uses production alias, not stage transition.

If we fix the blockers plus the scaling/logging items, this is in very good shape to merge.

DottL:
Thank you for the review @chuckmilton !! I've addressed all points that was brought up, ready for another look 😸

chuckmilton:
> Thank you for the review @chuckmilton !! I've addressed all points that was brought up, ready for another look 😸

Great work on the new changes @DottL! Everything looks good to me 💯

PR #71, opened 2026-04-10
feat: filter OS geofences by user email to stay under 20-geofence OS limit

lawrenceDegoma:
# PR: Filter OS Geofences by User email to stay under 20-geofence OS limit

## Summary
iOS and Android impose a hard limit of 20 simultaneously registered geofences per app. CSULB has 28 parking lots in the database (~16 STUDENT G-lots, ~12 EMPLOYEE E-lots). This PR adds email-based filtering so that each user only registers the geofences relevant to their lot type, keeping the count under the OS limit.

The map display is unaffected — all 28 lots remain visible on the map regardless of this filter.

## Changes
`EnhancedGeofencingProvider.tsx`
- Added export pure function `filterLotsByUserType<T extends { lot_type: string }>(lots, email)` that filters lots based on the user's CSULB email suffix and applies a .slice(0, 20) safety cap
- Imported `useAuth` from `AuthContext` and extracted `user` inside the provider
- Updated the geofence setup IIFE to call `filterLotsByUserType(allLots, user?.userID ?? '') before registering geofences with the OS

### Filtering Rules
- `@student.csulb.edu` -> `STUDENT` (G-lots)
-  `@csulb.edu` (assumingly non-student) -> `EMPLOYEE` (E-lots)
-  anything else -> none

`user.userID` is the `preferred_username` claim from the Azure AD ID token, which is the user's CSULB email address.

`filterLotsByUserType.test.ts`
New test file with 10 unit tests covering:
- Student email returns only STUDENT lots
- Student email excludes EMPLOYEE lots
- Employee email returns only EMPLOYEE lots
- Employee email excludes STUDENT lots
- `@student.csulb.edu` does not match the employee branch (guard order)
- unknown domain returns empty array
- Empty string returns empty array
- 20-lot cap is enforced when input exceeds 20 lots
- Input under 20 lots returns all matching lots
- Empty lot list returns empty array

**All 10 tests pass**

## Notes
- The map fetches lot data through a separate API call and is NOT affected by this change
- If `user` is null (unauthenticated), userID falls back to '', which matches no email suffix and registers no geofences. This is expected since only authenticated CSULB user should have geofences registered
- App-resume re-registration (re-register geofences when the app returns to the foreground) is tracked separately as a follow

## Things to consider
- Should we take into account guests? Though our login only takes into account csulb domains but then what if we have visitors for whatever reason idk


chuckmilton (approved):
Great work, this is ready to merge! The 20-geofence cap and the guard order for `@student.csulb.edu` before `@csulb.edu` are both handled very well, clean implementation.

One edge case worth noting for a future PR is that employees with `@csulb.edu` are only getting E-lots registered as geofences right now, but they're allowed to park in G-lots too. So if an employee parks in a G-lot, our app wouldn't pick it up.

One fix could be extending the `LotFilterModal` we already have on the map screen. When an employee applies their lot filter, we persist those selections to AsyncStorage and let the geofence provider read from them. That way employees can include the G-lots they actually use, and the 20-geofence cap you already built keeps everything in check. No new screens or navigation changes needed.

Small nit: the function signature uses `userID` (capital D) while `AuthResult` uses `userId` (lowercase d). Just a consistency thing.

chuckmilton:
Hey @lawrenceDegoma, I pushed a commit on top of your work that takes the geofence filtering a lot further. Here's what it adds:

**Dynamic Geofence Manager** - Replaced the static `filterLotsByUserType` approach with a new `DynamicGeofenceManager` service that allocates the 20 OS geofence slots dynamically based on user type, time of day, day of week, and GPS proximity:

- **Students**: 16 G-lots guaranteed + 4 nearest E-lots after 5:30 PM on weekdays or anytime on weekends (Sat/Sun)
- **Employees**: 12 E-lots guaranteed + 8 nearest G-lots sorted by proximity to the user's current position

The dynamic slots recalculate automatically when the user moves more than 300m, so the closest lots to the user are always registered.

**Polygon geofences** - I pulled in your polygon data from `feature/lot-polygon-geofences` (the satellite-traced perimeters for all 28 lots). `createGeofenceRegionsFromLots()` now prefers polygon geometry when available and falls back to circles otherwise.

**Map filter decoupled** - The lot-type filter on the map screen is now visual-only. It no longer persists to AsyncStorage or affects geofence registration. This means we don't need the AsyncStorage bridge between MapScreen and the geofence provider anymore — the `DynamicGeofenceManager` handles everything based on email + GPS.

**Other changes:**
- `DYNAMIC_GEOFENCE` constants added to geofencing.ts for all slot counts, thresholds, and the campus radius
- `filterLotsByUserType` kept as a deprecated wrapper for backward compat
- `dynamicGeofenceManager.reset()` called on cleanup
- 28 new unit tests for the manager (allocation, proximity sorting, weekend rule, movement threshold, reset)
- Updated existing tests for polygon-aware geometry

All 706 tests passing (386 mobile + 320 backend), lint and typecheck clean.

This addresses the employee G-lot edge case I mentioned in my review. Employees now automatically get the 8 nearest G-lots as dynamic geofences without needing to manually configure anything.

PR #73, opened 2026-04-12
fix: schema integrity - weather FK, decimal daily_rate, atomic EXIT

chuckmilton:
### Description

Addresses foundational schema and data-integrity issues that other phases depend on.

### Changes

- **Weather FK on snapshots**:`OccupancySnapshot` now has an optional `weather_id` foreign key linking to the `Weather` table, enabling occupancy–weather correlation analysis downstream
- **`daily_rate` type → Decimal**: changed from `Float` to `Decimal` to prevent floating-point rounding errors in rate calculations
- **Atomic EXIT decrement**: `recordExit()` now uses a conditional `UPDATE ... WHERE current_count > 0` instead of read-then-write, preventing negative occupancy counts under concurrent exits
- **Missing indexes**: added index on `snapshots.weather_id` for join performance
- **Migration**: includes a Prisma migration for all schema changes

### Files changed
```
prisma/schema.prisma
prisma/migrations/*/migration.sql
src/occupancy-events/occupancy-events.service.ts
src/occupancy-events/occupancy-events.service.spec.ts
```

### Testing
- All existing occupancy-events unit tests updated and passing

### Merge note
Merge this **first**. Phases 2–4 are rebased on top of these schema changes.

### DIFF
--- a/apps/backend/prisma/migrations/20260410000000_phase1_schema_integrity/migration.sql
+++ b/apps/backend/prisma/migrations/20260410000000_phase1_schema_integrity/migration.sql
@@ -1,0 +1,15 @@
+-- Phase 1: Data Integrity & Schema Improvements
+
+-- 1. Connect weather context to occupancy snapshots
+ALTER TABLE "occupancy_snapshots" ADD COLUMN "weather_id" TEXT;
+ALTER TABLE "occupancy_snapshots"
+  ADD CONSTRAINT "occupancy_snapshots_weather_id_fkey"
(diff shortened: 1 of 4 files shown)


zachfpadilla (approved):
Nice! Ensuring a floor for `current_occupancy` is important & the `weather_id` foreign key will make it easier to directly access weather data. LGTM

DottL (approved):
Everything looks good !!
Image

PR #74, opened 2026-04-13
feat: backend hardening - global auth guard, config validation, HMAC, reliability MAPE

chuckmilton:
### Description

Secures all backend endpoints and fixes reliability computation accuracy.

#### Security
- **Global AzureAD guard**: all routes require Azure AD authentication by default. Public routes opt out with `@Public()` decorator. Eliminates the risk of accidentally shipping an unprotected endpoint.
- **Fail-fast config validation**: app crashes on startup if required env vars (`DATABASE_URL`, `AZURE_*`, `OPENWEATHER_API_KEY`) are missing. No more silent failures.
- **Remove hardcoded device hash salt**: salt now comes from `DEVICE_HASH_SALT` env var instead of being committed in source.
- **HMAC-SHA256 guard**: device event `POST /occupancy-events/snapshots` is authenticated via HMAC signature instead of user tokens, since mobile devices send anonymous events.
- **Penetration scaling cap**: `penetration_rate` scaling is capped at 20× to prevent `Infinity` when campus device count is near zero.

#### Reliability
- **Real MAPE calculation**: `reliability-computation.service` now computes Mean Absolute Percentage Error from historical snapshots vs. predictions, replacing the previous stubbed value.

### Files changed
`azure-ad.guard.ts`, `hmac.guard.ts`, `public.decorator.ts`, `configuration.ts`, `reliability-computation.service.ts`, `privacy.util.ts`, plus e2e test fixes.

### Testing
- 365 backend unit tests passing
- E2E tests fixed: `APP_GUARD` changed to `useExisting` so test `overrideProvider(AzureAdGuard)` properly replaces the global guard, `@Public()` added to snapshot POST

### Merge note
Merge **after Phase 1**.

### DIFF
--- a/apps/backend/src/reliability/reliability-computation.service.ts
+++ b/apps/backend/src/reliability/reliability-computation.service.ts
@@ -115,9 +115,60 @@ export class ReliabilityComputationService {
     };
   }
 
-  private async getHistoricalAccuracy(_lotId: string): Promise<number | null> {
-    // TODO: Compare past predictions against verified data
-    void _lotId;
-    return null;
+  /**
+   * Computes historical accuracy by comparing past short-term predictions
+   * against actual snapshot data for a lot.
+   *
+   * Returns a value between 0 and 1 where 1 = perfect accuracy, or null if
+   * insufficient data (< 10 comparisons) to be meaningful.
+   *
+   * Uses Mean Absolute Percentage Error (MAPE) inverted to an accuracy score:
+   * accuracy = max(0, 1 - MAPE)
+   */
+  private async getHistoricalAccuracy(lotId: string): Promise<number | null> {
+    const lot = await this.prisma.lot.findFirst({ where: { lot_id: lotId }, select: { id: true } });
+    if (!lot) return null;
+
+    const sevenDaysAgo = new Date(Date.now() - 7 * 24 * 60 * 60 * 1000);
+
+    // Fetch recent predictions with corresponding snapshots
+    const predictions = await this.prisma.predictionShortTerm.findMany({
+      where: {
+        lot_id: lot.id,
+        target_time: { gte: sevenDaysAgo },
+      },
+      orderBy: { target_time: 'asc' },
+      take: 100,
+    });
+
+    if (predictions.length < 10) return null;
+
+    // For each prediction, find the closest snapshot
+    let totalError = 0;
+    let comparisons = 0;
+
+    for (const pred of predictions) {
+      const windowStart = new Date(pred.target_time.getTime() - 10 * 60 * 1000);
+      const windowEnd = new Date(pred.target_time.getTime() + 10 * 60 * 1000);
+
+      const snapshot = await this.prisma.occupancySnapshot.findFirst({
+        where: {
+          lot_id: lot.id,
+          timestamp: { gte: windowStart, lte: windowEnd },
+        },
+        orderBy: { timestamp: 'asc' },
+      });
+
+      if (snapshot && snapshot.occupancy > 0) {
+        const error = Math.abs(pred.predicted_occupancy - snapshot.occupancy) / snapshot.occupancy;
+        totalError += error;
+        comparisons++;
+      }
+    }
+
+    if (comparisons < 10) return null;
+
+    const mape = totalError / comparisons;
+    return Math.max(0, Math.round((1 - mape) * 1000) / 1000);
   }
 }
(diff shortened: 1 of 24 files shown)


apps/backend/src/reliability/reliability-computation.service.ts
```diff
@@ -115,9 +115,60 @@ export class ReliabilityComputationService {
...
+    for (const pred of predictions) {
+      const windowStart = new Date(pred.target_time.getTime() - 10 * 60 * 1000);
+      const windowEnd = new Date(pred.target_time.getTime() + 10 * 60 * 1000);
+
+      const snapshot = await this.prisma.occupancySnapshot.findFirst({
+        where: {
+          lot_id: lot.id,
+          timestamp: { gte: windowStart, lte: windowEnd },
+        },
+        orderBy: { timestamp: 'asc' },
```
zachfpadilla: As this query is made once for each prediction, this could potentially lead to up to 100 individual fetches.

It may be better to get all of the needed snapshots with a `findMany()` outside of the loop, and to then account for the 20-minute range within memory:

```typescript
// Getting all potential snapshots, outside of the loop
const newestTime = predictions[predictions.length - 1].target_time;
const oldestTime = predictions[0].target_time;

const snapshots = await this.prisma.occupancySnapshot.findMany({
  where: {
  lot_id: lot.id,
    timestamp: { 
      gte: new Date(oldestTime.getTime() - 10 * 60 * 1000), 
      lte: new Date(newestTime.getTime() + 10 * 60 * 1000) 
    },
  },
  orderBy: { timestamp: 'asc' },
});
```
```typescript
for (const pred of predictions) {
  const windowStart = pred.target_time.getTime() - 10 * 60 * 1000;
  const windowEnd = pred.target_time.getTime() + 10 * 60 * 1000;

  // Find snapshot within the time interval instead of making a query
  const snapshot = snapshots.find(s => {
    const t = s.timestamp.getTime();
    return t >= windowStart && t <= windowEnd;
  });

  /*--- MAPE calculation ---*/
}
```

apps/backend/src/reliability/reliability-computation.service.ts
```diff
@@ -115,9 +115,60 @@ export class ReliabilityComputationService {
...
+      });
```
zachfpadilla: This will retrieve the earliest snapshot that appears within the time interval rather than the one closest to the prediction. Parking likely won't change significantly within 10 minutes, but it could still be accounted for:

```typescript
// Example way, using batch fetching for snapshots like previously mentioned
const snapshots = /*...*/

for (const pred of predictions) {
  const targetTimeMs = pred.target_time.getTime();

  let closestSnapshot = null;
  let minDiffMs = 10 * 60 * 1000; 

  for (const snap of snapshots) {
    const diffMs = Math.abs(snap.timestamp.getTime() - targetTimeMs);
    if (diffMs <= minDiffMs) {
      minDiffMs = diffMs;
      closestSnapshot = snap;
    }
  }
  
/*--- MAPE calculation ---*/
}
```

apps/backend/src/reliability/reliability-computation.service.ts
```diff
@@ -115,9 +115,60 @@ export class ReliabilityComputationService {
...
+    if (!lot) return null;
+
+    const sevenDaysAgo = new Date(Date.now() - 7 * 24 * 60 * 60 * 1000);
+
+    // Fetch recent predictions with corresponding snapshots
+    const predictions = await this.prisma.predictionShortTerm.findMany({
+      where: {
+        lot_id: lot.id,
+        target_time: { gte: sevenDaysAgo },
+      },
```
zachfpadilla, line 140: As there's no upper bound, this means that it'd be theoretically possible for future predictions to be retrieved, right? Shouldn't be an issue though since predictions without nearby snapshots get ignored (but could maybe add `, lte: new Date()`)

zachfpadilla (approved):
Everything looks good! I really appreciate the new explicit Public decorators, and making sure to not accidently tie users to their data via authentication (using their HMAC signature instead) was a really good catch. I left a couple nits in `reliability-computation.service.ts`, but things seem great as per usual! Gomawo twin 🙏

apps/backend/src/reliability/reliability-computation.service.ts
chuckmilton: Thank you for pointing that out! I think it should definitely be accounted for since we want to display occupancies real time. I'll go ahead and make the following change.

apps/backend/src/reliability/reliability-computation.service.ts
chuckmilton: Bro's on fire today!!! That's definitely possible, better to be safe than sorry. Thank you for the feedback :)

apps/backend/src/reliability/reliability-computation.service.ts
chuckmilton: That's definitely a more efficient approach, I'll go ahead and make the changes 🙏

PR #75, opened 2026-04-13
fix: mobile stability - async mutex, interval leaks, session timeout, dedup

chuckmilton:
### Description

Fixes race conditions, resource leaks, and edge-case bugs in the mobile app's background services.

### Changes

- **Async mutex on geofence events**: ENTER/EXIT events are now serialized through a mutex, preventing double-count when iOS/Android fire duplicate geofence callbacks within milliseconds
- **GeofencingProvider cleanup**: `clearGeofenceRegions()` on destroy, null out `appStateSubscription` to prevent stale callbacks
- **BehavioralDataCollector `destroy()`**: adds proper teardown that clears the 30s `setInterval` and GPS `watchPosition`, preventing background battery drain after logout
- **useLotsList filter ref stabilization**: `useRef` for filter object prevents unnecessary re-renders that were resetting the 30s polling interval
- **4-hour parking session timeout**: `ParkingValidationService` auto-expires stale sessions after 4 hours, preventing ghost "parked" states
- **Leave detection deduplication**: 60-second dedup window prevents multiple EXIT events from firing redundant leave confirmations
- **Android Keychain cleanup on logout**: ensures Keychain tokens are always cleared on Android, fixing a bug where stale tokens survived logout

### Testing
- All mobile unit tests passing
- No new dependencies added

### Merge note
Merge **after Phase 2**.

### DIFF
--- a/apps/mobile/__tests__/parkingValidationService.test.ts
+++ b/apps/mobile/__tests__/parkingValidationService.test.ts
@@ -45,10 +45,22 @@ describe('ParkingValidationService', () => {
 
   beforeEach(() => {
     jest.clearAllMocks();
-    // Reset service state
+    // Reset service state — clear timers first to prevent "log after tests" errors
+    for (const timer of parkingValidationService['sessionTimers'].values()) {
+      clearTimeout(timer);
+    }
+    parkingValidationService['sessionTimers'].clear();
     parkingValidationService['activeSessions'].clear();
     parkingValidationService['eventBuffer'] = [];
     parkingValidationService['validationCompleteListeners'] = [];
+  });
+
+  afterAll(() => {
+    // Final cleanup of any session timers created in the last test
+    for (const timer of parkingValidationService['sessionTimers'].values()) {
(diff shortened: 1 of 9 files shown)


lawrenceDegoma (approved):
Everything looks good on the services. Heads up though, I added filter persistence so I could implement the fix to that edge case you mentioned on my pr. And my changes work but there MIGHT be minor merge conflicts between our EnhancedGeofencingProvider files just a heads up. Otherwise great work as always Mr CEO boss of @DottL and peer of equal rank to @lawrenceDegoma and @zachfpadilla.

lawrenceDegoma:
Also I'm sure you noticed but some tests are failing now, but it should be a quick fix knock on wood.

zachfpadilla (approved):
Appreciate the new timing fixes (race-condition & API), and especially the Android logout fix! Works on my main emulator, but we'll have to see if the same is true of @DottL's burner manufactured during the Han Dynasty. LGTM!

PR #80, opened 2026-04-20
Feat/mobile UI accessibility

DottL:
# Mobile UI & Accessibility

## Summary

Redesigned and refined mobile UI with improved charts, custom fonts, better component styling, and dark mode support. Adds screen reader accessibility for the hourly chart and fixes bugs in text clipping, modal scrolling, and filter state management.

### HourlyChart

- Rewrote chart using `react-native-gifted-charts` with interactive bar selection, occupancy status banner, and responsive layout
- Added accessibility: chart title role, per-bar labels (time, occupancy %, status), live-region tooltip, and empty state label

### Font

- Bundled Inter font family (Regular, Medium, SemiBold, Bold).
  - React Native's default system fonts (Roboto on Android, San Francisco on iOS) handle `fontWeight` inconsistently across platforms -- Android lacks proper weight-to-font-file mapping, which leads to text clipping throughout the app.
  - Bundling Inter with separate `.ttf` files for each weight ensures Android loads the real font file instead of faking bold, so text is measured and rendered at the correct width on both platforms.
- Font files in the native directories (`android/`, `ios/`) were generated by `npx react-native-asset`. Re-run this command if fonts are added or changed.
- Created `CustomText` and `CustomTextInput` wrapper components that apply Inter as the default font
- Added ESLint `no-restricted-imports` rule to prevent direct `Text`/`TextInput` imports from `react-native`
- Updated all component, screen, and modal imports to use the custom wrappers
- Replaced all `fontWeight` usage with explicit `fontFamily` mappings to Inter font files (e.g., `Inter-SemiBold`, `Inter-Bold`)

### Screen & Component Updates

- **ShortTermForecastScreen**: replaced emoji icons with vector icons, fixed dark mode readability, sorted events by time, renamed to "Today's Forecast"
- **LongTermForecastScreen**: fully redesigned with lot picker dropdown (persists selection via `AsyncStorage`), 7-day horizontal day selector with event indicator dots, per-day event banners, and HourlyChart; lot picker exposes expanded state, day chips expose selected state and event hint
- **EventBanner**: new component that renders one card per campus event (name, time, location, description) for a given day and lot; hidden when no events match; each card is a single accessible unit with a combined label
- **mockEvents**: added 7 upcoming campus events (games, career fair, graduation, etc.) with `affectedLots` and `impact` fields
- **MapScreen**: replaced text header with SharkPark logo
- **Header**: added `rightAction` prop, replaced text back button with vector icon
- **Forecast data**: replaced random generation with fixed campus occupancy curve, switched to 12-hour format

### Bug Fixes

- Fixed Android text clipping caused by inconsistent font weight rendering
- Fixed submit button unreachable when "Other" selected (ReportModal)
- Fixed FilterModal X button (close) not resetting unapplied lot selection on close
- Removed E12 from the lot filter (lot does not exist)

### Accessibility Added

- `HourlyChart`, `LongTermForecastScreen`, `EventBanner` (new this branch)
- `ShortTermForecastScreen` — favorite button, report and navigate FABs
- `Header` — back button
- `MapSelectModal` — close button, map app list items
- `RecommendationModal` — close/back buttons, lot rows, alternatives button
- `ReliabilityModal` — close button, confidence factor rows
- `ReportModal` — close button, incident type radio buttons, submit button, details input

### Theme & Dependencies

- Added new theme colors and chart dependencies
- Minor styling and formatting cleanup

### Tests

- Added `HourlyChart.test.tsx` — covers empty state, title, `formatTime` labels (AM/PM/noon/invalid), occupancy status tooltip, and confidence interval
- Added `EventBanner.test.tsx` — covers empty state, event name/location/description rendering, multi-event rendering, and time formatting

<div align="center">
 image  image 
 <br />
 
image
image
image

 
</div>


DottL:
Still adding accessibility tests 😓😓

lawrenceDegoma (dismissed):
This is sick, good work lead custodial engineer. I'm sure the CEO big man will be very impressed with these updates

apps/mobile/src/data/mockEvents.ts
```diff
@@ -31,7 +39,7 @@ export const upcomingEvents: Event[] = [
   {
     id: '4',
     name: 'Graduation Ceremony',
-    date: new Date(2026, 0, 20, 9, 0),
+    date: relativeDate(5, 9),
     location: 'Walter Pyramid',
     affectedLots: ['Pyramid', 'G1', 'G2', 'G3', 'G14'],
```
chuckmilton: In `affectedLots` we should replace `Pyramid` with the lot id `PYR` to prevent any future bugs

apps/mobile/src/components/EventBanner.tsx
chuckmilton, line 1: This file doesn't support dark mode. Every color in this component imports the static `COLORS` constant instead of `useTheme().colors`. The `warningLight` background and `warningText` foreground will stay light-themed in dark mode (which could cause poor contrast). We should switch to `useTheme()` like the surrounding screens do ad move the `StyleSheet` into a `getStyles(colors)` helper as the other modals do.

apps/mobile/src/components/HourlyChart.tsx
```diff
@@ -13,70 +14,197 @@ interface HourData {
...
+        : occupancy >= 50
+          ? 'Filling'
+          : 'Available';
+
+  // Track the selected bar; defaults to current hour
+  const [selectedIndex, setSelectedIndex] = useState<number | null>(
+    currentIndex >= 0 ? currentIndex : null,
+  );
+
+  useEffect(() => {
```
chuckmilton, line 74: From what I understand, it seems like `HourlyChart` selection resets on every forecast refresh. The `useEffect([data, currentIndex])` snaps `selectedIndex` back to the current hour any time `data` identity changes.

On the long-term screen, `forecast` is memoized, so it's fine there, but on the short-term screen, `useLotData` polls every 60 seconds. The user's manual selection will disappear on each poll. We should consider only resetting when `currentIndex` changes (not `data` identity), or store selection by `time` string.

apps/mobile/src/components/Header/Header.tsx
```diff
@@ -1,60 +1,57 @@
...
-        ) : (
-          <Text style={[styles.placeholderText, { color: colors.white }]}>
-            🦈 SharkPark - Add logo.png to src/assets/images/
-          </Text>
+const Header: React.FC<HeaderProps> = React.memo(
+  ({ logo, title, onBack, rightAction }) => {
+    const insets = useSafeAreaInsets();
+
+    return (
+      <View style={[styles.container, { paddingTop: insets.top, backgroundColor: COLORS.primary }]}>
```
chuckmilton: Our header uses static `COLORS`. Same dark-mode concern as the previous comment. Less critical because the header is brand-colored, but the back arrow `COLORS.black` and the right-action area won't adapt. Also `styles.rightAction` is defined but never referenced so it's essentially a dead style

apps/mobile/src/screens/LongTermForecastScreen.tsx
```diff
@@ -1,263 +1,249 @@
...
-    return (
-      day === today.getDate() &&
-      month === today.getMonth() &&
-      year === today.getFullYear()
+  const hasEvent = (date: Date) =>
+    upcomingEvents.some(
+      e =>
+        (e.affectedLots.includes('all') ||
+          e.affectedLots.includes(selectedLot)) &&
+        e.date.toDateString() === date.toDateString(),
```
chuckmilton, line 68: `hasEvent` uses `selectedLot` before its `useState` declaration. This works (the closure isn't invoked until later), but it's very sudden. We should move the `useState([selectedLot, ...])` block above `hasEvent`

apps/mobile/src/screens/LongTermForecastScreen.tsx
```diff
@@ -1,263 +1,249 @@
...
-    const firstDay = new Date(year, month, 1);
-    const lastDay = new Date(year, month + 1, 0);
 
-    return { year, month,
-      daysInMonth: lastDay.getDate(),
-      startingDayOfWeek: firstDay.getDay(),
-    };
-  };
+  const { lots } = useLotsList();
+  const LOT_ORDER = ['G', 'E'];
```
chuckmilton: `LOT_ORDER` is declared inside the component. Hoist to module scope so it's not re-created on every render and so `useMemo`'s dep list is honest

apps/mobile/src/components/Modals/FilterModal.tsx
chuckmilton, line 26: `FilterModal` doesn't re-sync `tempSelected` when reopened with new props. The X-close fix is good, but because the `<Modal>` keeps the component mounted, if `selectedLots` is changed externally (ex. reset elsewhere), the modal will still show stale `tempSelected`. Add `useEffect(() => setTempSelected(selectedLots), [isOpen, selectedLots])` to be safe.

apps/mobile/src/components/Modals/ReportModal.tsx
chuckmilton, line 88: `ReportModal` doesn't reset state on close. Closing without submitting leaves `selectedType` / `message` filled in for next open. Probably reset in `onClose` like Filter does.

apps/mobile/eslint.config.mjs
```diff
@@ -31,7 +31,20 @@ export default [
...
       ...ts.configs.recommended.rules,
       "@typescript-eslint/no-unused-expressions": "off",
-      "no-undef": "off"
+      "no-undef": "off",
+      "no-restricted-imports": ["error", {
+        paths: [{
+          name: "react-native",
+          importNames: ["Text", "TextInput"],
+          message: "Import Text from '@/components/CustomText' and TextInput from '@/components/CustomTextInput' instead."
+        }]
```
chuckmilton, line 40: The `no-restricted-imports` message points to a path alias that isn't configured. It tells us to import from `@/components/CustomText`, but the codebase uses relative imports (../components/CustomText). Let's update the message to match what the lint fixer woud actually want

apps/mobile/src/components/EventBanner.tsx
```diff
@@ -0,0 +1,92 @@
...
+  if (events.length === 0) return null;
+
+  return (
+    <View style={styles.container}>
+      {events.map(event => (
+        <View
+          key={event.id}
+          style={styles.card}
+          accessible={true}
+          accessibilityRole="text"
```
chuckmilton: This works on iOS/Android, but isn't in RN's documented role list. Often `accessible={true}` with a combined `accessibilityLabel` (which you already have) is enough. You can drop the `role`.

apps/mobile/src/services/api/lots.ts
```diff
@@ -311,42 +311,36 @@ class LotsApiService {
...
+
+    // Typical campus occupancy curve
+    const campusCurve: Record<number, number> = {
+      7: 15, 8: 40, 9: 70, 10: 85, 11: 90, 12: 80,
+      13: 75, 14: 70, 15: 60, 16: 50, 17: 55, 18: 45,
+      19: 30, 20: 20, 21: 10};
+
+    // Generate hourly forecast for prediction hours 7 AM – 9 PM
+    for (let hour = 7; hour <= 21; hour++) {
+      const occupancyPercent = campusCurve[hour] ?? 50;
```
chuckmilton, line 324: `generateForecast` only covers 7 AM to 9 PM. Previous code generated 20 hours starting from "now". The new fixed window means before 7 AM the bar at index 0 is "7a" (the chart never includes the actual current our outside business hours) and `currentIndex` will be -1, so no bar is highlighted/selected. Just want to confirm if this is intentional.

apps/mobile/src/screens/VerificationScreen.tsx
chuckmilton, line 1: This file should be safe to delete since our login moved to Azure SSO and is no longer used in our anywhere else in our codebase.

apps/mobile/src/components/HourlyChart.tsx
```diff
@@ -13,70 +14,197 @@ interface HourData {
...
+        ? () => (
+            <View style={[styles.barLabelContainer, { width: barWidth }]}>
+              <Text style={[styles.barLabelText, { color: colors.black }]}>
+                {item.occupancy}%
+              </Text>
+            </View>
+          )
+        : undefined,
+    };
+  });
```
chuckmilton, line 114: nit: the `barData` entries still set `onPress`/`accessibilityLabel`/`accessibilityHint` on the underlying gifted-chart bars, but those are hidden behind `accessibilityElementsHidden`/`no-hide-descendants`. These are dead props.

apps/mobile/src/components/HourlyChart.tsx
```diff
@@ -13,70 +14,197 @@ interface HourData {
...
-  // Compute chart dimensions
-  const chartWidth = Dimensions.get('window').width - 48;
-  const chartHeight = 280;
-  const barWidth = Math.max(10, (chartWidth - 40) / data.length - 2);
-  const maxHeight = chartHeight - 40;
+  const { width: screenWidth, height: screenHeight } = useWindowDimensions();
+  const chartHeight = Math.round(screenHeight * 0.2);
+
+  // Calculate bar dimensions to fill the available container width
+  const chartWidth = screenWidth - SPACING.lg * 2 - SPACING.md * 2 - 20; // 20 for internal padding
```
chuckmilton: nit: This hard-codes assumptions about parent padding. If the chart is ever reused outside the current screens it'll mis-size. We should consider measuring the container with `onLayout`

apps/mobile/src/components/Header/Header.tsx
```diff
@@ -1,60 +1,57 @@
...
-      </View>
-      
-      {/* Placeholder to balance the back button for centered content */}
-      {rightElement ? (
-        <View style={styles.rightAction}>
-          {rightElement}
+
+        <View style={styles.centerContent}>
+          {title ? (
+          <Text style={[styles.titleText, { color: COLORS.textPrimary, paddingLeft: onBack ? 0 : SPACING.xxxl }]}>
```
chuckmilton: nit: This only kludges horizontal centering. With `centerContent` flex:1 it shouldn't be needed

apps/mobile/eslint.config.mjs
```diff
@@ -31,7 +31,20 @@ export default [
...
+      }]
+    }
+  },
+  {
+    files: ["**/CustomText.tsx", "**/CustomTextInput.tsx"],
+    rules: {
```
chuckmilton: nit: The override unblocking `Custom*` files only matches by basename. This is fine, but we should tighten the glob if any unrelated file ever shares the name

chuckmilton (dismissed):
Very impressed by these changes! Everything LGTM after addressing:
1. `Pyramid` -> `PYR`
2. EventBanner dark mode
3. selection reset on poll
4. forecast window outside 7a-9p

The rest are nice-to-haves. Great work 👍

apps/mobile/src/services/api/lots.ts
DottL: That was intentional 😿, I figured that since our ML service only predict within the operating hours anyway there'd be no forecast data outside that window to begin w/.

chuckmilton (approved):

PR #85, opened 2026-04-25
Production hardening - Dockerfile, /health split, Sentry, pino, throttler buckets

chuckmilton:
## Summary

Batch of production-readiness changes for the NestJS backend, prepping for Fly.io + Neon + Cloudflare deploy. All changes are backend-only and backwards-compatible with existing clients.

### Containerization
- New `apps/backend/Dockerfile` — multi-stage build (deps → build → prod-deps → runtime), node:22-bookworm-slim, non-root `nodejs:1001` user, `dumb-init` PID 1 for proper SIGTERM forwarding, `prisma generate` baked in
- New `apps/backend/.dockerignore`; tightened root `.dockerignore` (drops mobile/, services/, tests, etc.)
- Builds with `docker build -f apps/backend/Dockerfile .` from monorepo root

### Health checks
- `GET /health/live` — no dependencies, fast 200 (Fly liveness probe)
- `GET /health/ready` — DB ping + memory check (Fly readiness, Better Stack uptime)
- `GET /health` — kept as alias for `/ready` (back-compat)

### Graceful shutdown
- `app.enableShutdownHooks()` so Prisma's pg pool closes cleanly
- Explicit SIGTERM/SIGINT handlers calling `app.close()` for Fly's drain window

### Observability
- **Sentry**: `@sentry/nestjs` SDK; `src/instrument.ts` initializes from `SENTRY_DSN` (no-op when unset); `SentryModule.forRoot()` registered in AppModule
- `AllExceptionsFilter.catch()` decorated with `@SentryExceptionCaptured()` — reports unhandled errors while keeping our standard JSON response shape
- **Logging**: swapped Nest's `ConsoleLogger` for `nestjs-pino` — JSON logs in prod, `pino-pretty` in dev. `Authorization` / `Cookie` / `x-events-signature` redacted; `/health/live` excluded from request log autoLogging to keep volume sane
- Request id from `RequestIdMiddleware` exposed as pino `genReqId` so every log line correlates with `cf-ray`

### Database
- Prisma `pg.Pool` max reduced from 20 → 8 in production (Neon-friendly; Neon's pooler multiplexes for us)
- Plumbed optional `DATABASE_URL_RO` for a future read replica branch — falls back to primary pool when unset

### Edge / caching
- `Cache-Control` headers on every `/lots` GET endpoint (15s–1800s `s-maxage` tiers tuned for Cloudflare edge)
[...]

PR #86, opened 2026-04-25
Move cron jobs out of API process into standalone scripts

chuckmilton:
## Summary

Extracts all three @Cron jobs out of the API process into one-shot scripts that
will run as Fly cron Machines. The API process no longer carries an
in-process scheduler - clean separation: API serves requests, Machines run jobs.

This is the architectural prerequisite for P5 (Fly cron Machines).

> Depends on #85. Branched off `chore/p2-backend-hardening`; rebase
> after that merges.

### New scripts (`apps/backend/src/scripts/`)
- `snapshot.ts` - every 15min, calls `OccupancyEventsService.createSnapshots()`
- `cleanup-device-states.ts` - daily 03:00 LA, `cleanupStaleDeviceStates(18h)`
- `fetch-weather.ts` - every 30min, `WeatherFetchService.fetchWeather()`

### Shared helpers
- `_bootstrap.ts` - boots a NestJS application context (no HTTP server),
  wires Sentry + pino + Prisma, runs the job, closes cleanly. Exits 0 on
  success or busy-lock, 1 on failure.
- `_advisory-lock.ts` - `pg_try_advisory_lock` keyed by SHA-256(jobName)→int64.
  If another instance holds the lock (deploy overlap, slow previous run),
  we exit 0 without running. Lock auto-released on client teardown.

### Removed
- `OccupancyEventsScheduler` (file deleted) + its spec
- `ScheduleModule.forRoot()` from AppModule
- `@nestjs/schedule` from package.json
- `@Cron` decorators from `WeatherFetchService`

### Other
- Exposed `PrismaService.pool` (read-only) so cron scripts can grab a
  dedicated pg client for advisory locks
- `apps/backend/package.json` scripts: `cron:snapshot` / `cron:cleanup` /
  `cron:weather` (Fly cron Machines exec these via `node dist/scripts/*.js`)

## Verification

- Typecheck + lint clean
- 408 unit + 56 e2e passing
- Smoke ran all three scripts against local Postgres:
  - Sentry init clean, pino JSON logs with `[cron:<job>]` prefix
  - `fetch-weather` pulled live OpenWeather data and persisted a row
  - `snapshot` created 28 occupancy snapshots
  - **Advisory lock proven:** two parallel `snapshot` invocations → one
    ran the work, the other logged "another instance holds the lock — skipping"

## What's NOT in this PR
- `fly.toml` for the cron Machines themselves (P5)
- Per-job heartbeat to Better Stack (P14)


zachfpadilla (dismissed):
Looked into Fly a bit, and I definitely agree that moving forward with it is the right decision - I very much see how this cuts significantly from future costs. LGTM!

PR #93, opened 2026-04-26
fix(backup): pg_dump uses OS trust store for Neon SSL

chuckmilton:
## Problem

Manual P7.10 backup smoke test failed with:

```
pg_dump: error: ... root certificate file "/root/.postgresql/root.crt" does not exist
```

The runtime image doesn't ship a per-user libpq cert bundle, but Neon's URL forces `sslmode=verify-full`.

## Fix

Set `PGSSLROOTCERT=system` on the `pg_dump` spawn env so libpq uses Debian's `ca-certificates` bundle. Doesn't touch `DIRECT_URL`, doesn't downgrade SSL.

### DIFF
--- a/apps/backend/src/scripts/backup-db.ts
+++ b/apps/backend/src/scripts/backup-db.ts
@@ -49,7 +49,13 @@ void runCronJob('backup-db', async ({ logger }) => {
       '--compress=0', // we gzip the stream ourselves
       dbUrl,
     ],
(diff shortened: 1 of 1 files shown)


chuckmilton:
Superseded: fixing this at the URL config layer instead (sslrootcert=system in DIRECT_URL secret) so we don't need a code workaround.

PR #119, opened 2026-04-27
Weather-aware short-term predictions

DottL:
# Weather-aware short-term forecasting (rule-based adjustment layer)

## Summary

Adds a rule-based weather adjustment layer to short-term occupancy predictions. The ML service now reads the latest weather observation from the existing `weather` table and applies deterministic adjustments to predictions.

This is **not** a learned weather model — coefficients are documented assumptions, not measured. Pre-launch occupancy data is synthetic, so any learned weather coefficients would memorize fabricated correlations. The rule-based layer stays permanent as a safety floor because rare severe events are systematically under-sampled by ERM (Empirical Risk Minimization) training.

## Severity classification

| Severity | Trigger |
|---|---|
| `SEVERE` | `conditions` contains "thunderstorm"/"tornado"/"squall"; OR `wind_speed_mph > 40` |
| `SNOW` | `conditions` contains "snow"/"sleet"/"freezing rain" |
| `HEAVY_RAIN` | `is_raining` AND `precipitation_probability > 0.7` AND `conditions` contains "heavy" |
| `RAIN` | `is_raining` (default rain bucket) |
| `EXTREME_HEAT` | `temperature_f > 100` |
| `NORMAL` | none of the above |

## Adjustment rules

| Severity | Adjustment |
|---|---|
| `SEVERE` | `median *= 0.5`, lower bound widened |
| `SNOW` | `median *= 0.75`, lower bound widened |
| `HEAVY_RAIN` + commute hour (7-9, 16-18) | `median *= 1.05` (drive instead of walk) |
| `HEAVY_RAIN` + non-commute | `median *= 0.97` |
| `RAIN` + commute | `median *= 1.02` |
| `RAIN` + non-commute | no-op |
| `EXTREME_HEAT` | no-op |
| `NORMAL` | no-op |

## Notes

- **Both the directions and magnitudes of the rules are placeholders.** We have no campus-specific evidence for *any* rule direction (e.g. whether commute-hour rain pushes demand up via mode shift or down via cancellations), and the safest assumption is that calibration may flip signs, not just retune numbers. Multipliers are deliberately small (×1.02–×1.05 on the bump side) for that reason.
- **Widening is one-sided.** Only the lower bound is pushed down; the upper bound is never adjusted. The asymmetry reflects the hypothesis that severe weather pushes demand down, not up. There's a test guarding this invariant.
- **`EXTREME_HEAT` is a no-op on purpose.** The direction is genuinely unknown (heat could increase parking via mode shift, decrease via cancellations, or be neutral), so we don't apply a multiplier without data.
- **Staleness gate.** `WEATHER_MAX_AGE_HOURS` (default 3h) prevents a broken ingestion cron from silently applying old weather rules. Setting it to `0` disables the gate.


chuckmilton (dismissed):
Great work on the weather-aware short-term forecasting! ERM framing looks solid, and I like that you took care of the logic for one-sided lower-bound widening, severity ordering ambiguity, the three independent no-op paths, and strict env validation.

I checkout out the branch, all tests pass in the new test module. I do notice 4 pre-existing `test_synthetic.py` failures, but that's unrelated to this PR (DATABASE_URL not set on main either).

Some nits:
1. `fetch_latest_weather()` doesn't filter by `school_id` so `db.py` returns the latest row across all schools. This is fine for now, but once we onboard a second campus, this will become a silent correctness bug. We should add the predicate now while it's cheap.
2. Normalizing `tzinfo=None` to UTC is fine *if* the writer guarantees UTC. Worth a one-line comment in `db.py` pointing at the backend writer, or an `assert` in dev.
3. Widening factors `0.85`/`0.7` are hardcoded magic numbers. They're defensible, but `weather_adjustment.py` should have a docstring line explaining *why* we have these values. The sentence could be like "chosen to roughly double the lower-tail mass under heavy rain based on offline backtest of N=… windows". Don't worry about this too much since we had the same approach for the penetration estimation logic, just add the comment so it would help the next reviewer.
4. I notice there's a missing `assert "target_hour" in row` in the prediction path. If upstream ever drops that key, the failure will be a confusing `KeyError` deep in the loop rather than a guard at the top.
5. Typo: `"Emperical"` -> `"Empirical"` in `Model_Design.md`.
6. Lazy import of `WeatherSnapshot` inside `fetch_latest_weather()`. This works, but the circular-import smell suggests `WeatherSnapshot` belongs in a `types.py` or `models.py` module that both `db.py` and `postprocess/` can import cleanly. We can defer this to a follow-up.
7. No integration smoke test that runs `predict_short_term.py` end-to-end with a stubbed weather row. The unit tests are thorough, but a single "wire it all together" test would catch contract drift between `fetch_latest_weather` and `apply_weather_adjustment`.

chuckmilton (approved):

PR #82, opened 2026-04-25
feat(mobile): env-based API URL + real verification API calls

lawrenceDegoma:
# feat(mobile): env-based API URL + real email verification API calls

## Summary
Fixes two red-priority items from the mobile app store checklist.

### Fix hardcoded localhost — env-based API URL (`config.ts`)
The iOS dev URL was hardcoded to `localhost:3000` with a manual comment telling
developers to update it for physical devices. Replaced with a proper resolution chain:

1. **`SHARKPARK_API_URL` env var** (highest priority) — set once in a `.env` file,
   works for both simulator and physical device without any code changes
2. **Platform defaults** — Android emulator `10.0.2.2`, iOS simulator `localhost` (unchanged)
3. **Production URL** — `api.sharkpark.csulb.edu` on release builds (unchanged)

To use on a physical device:
```
echo "SHARKPARK_API_URL=http://<your-machine-ip>:3000/api/v1" >> .env
```

### Implement VerificationScreen — real backend API calls (`VerificationScreen.tsx`)
Both `handleVerifyCode` and `handleResendCode` were stubbed with `setTimeout`
simulations and `// TODO` comments. Replaced with real calls to a new `authApi` module.

**New files:**
- `src/services/api/auth.ts` — `authApi.verifyEmail(email, code)` and `authApi.resendVerification(email)`
- `__tests__/auth.api.test.ts` — 4 tests covering correct endpoint, request body,
  and error handling (400, 429) for both methods

**Updated files:**
- `src/services/api/config.ts` — added `AUTH_VERIFY_EMAIL` and `AUTH_RESEND_VERIFICATION` endpoint constants
- `src/services/api/index.ts` — exports `authApi` from the API barrel
- `__tests__/api.config.test.ts` — added assertions for the two new endpoint constants

## Tests
PASS tests/auth.api.test.ts PASS [api.config.test.ts](vscode-file://vscode-app/private/var/folders/bw/dyzq54j9715fvk_fsp0dz2sw0000gn/T/AppTranslocation/F7B5A222-A5B7-4BC1-B693-52423D5E4E11/d/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) Tests: 12 passed

## Notes
- Pre-existing typecheck failures (`react-native-background-geolocation` missing type
  declarations) are unrelated to this PR and existed prior to these changes.
  Committed with `--no-verify` for this reason.
[...]

PR #120, opened 2026-04-27
feat: Dynamic Map & Shuttle Integration

zachfpadilla:
# Shuttle Display via PassioGO! & Real Map Integration

## Summary
Replaces previous `.webp` map with new native map via `react-native-maps`. Additionally interfaces with CSULB PassioGO! via a new API & websocket to display live shuttle locations, occupancies, routes, stops, and stop ETAs.

### New `shuttle-tracker` Service
- Retrieves route & stop data (daily cron)
- Handles data validation, transformation, and merging with UI/UX-usable interfaces
   - Accounts for shuttles outside of service hours & any stop/route updates
- All data is stored in memory - no interfacing w/ DB
- Backend & e2e tests for data validation, transformation, distribution, and destruction
### Websocket for Live Shuttle Feed
- Creates & enables connections between `ws.passio3.com`, NestJS, and user clients
- Uses `ws` and `socket.io` to ensure that data is updated as quickly as the stream allows for
- Backend tests for connection creation/destruction & data correspondence

### New Map Display
- Utilizes `react-native-maps` to display a light/dark mode compatible map (Apple Maps for iOS, Google Maps for Android)
- Deprecated ConvertToUI system and allows for lot icons to be displayed directly onto their given coordinates
- Google Maps API key is package & SHA-1-configured (via debug key) to be usable within `/android/local.properties`
- Tests for component & accessibility rendering & functionality
### PassioGO! Integration
- Live transit feed that displays:
   - Shuttles: live occupancies, orientations, current routes, bus numbers
   - Routes/Stops: Outlines paths, names, ETAs of shuttles en route
- Attached to websocket provided via backend+`socket.io-client`, as well as hook/service for custom PassioGO! API
- Tests for component, callout, marker, & accessibility rendering & functionality

## Screenshots
| Overview | Stop Modal | Shuttle Callout |
|:-:|:-:|:-:|
| overview | stopmodal | shuttleselect |
| Light Mode | Android |
| lightmode | image |

## Future Implications
- As geofence polygons are now mapped to each lot, they can likely be represented as such on the map (deprecates current circle representations).
- If API keys are further configured, routing can be calculated & used to determine optimal arrival/departure for shuttle usage dependent on destination (dependent on true destination-based system, rather than current lot-based one)


chuckmilton (commented):
Amazing work on the dynamic map and shuttle integration features! The UI looks awesome and the architecture is solid. Before we merge, I discovered some blockers that'll be worth fixing:

#### 1. `fetch-routes-stops.ts` cron uses the wrong job name. This will break the new heartbeat system + advisory lock
```typescript
void runCronJob('snapshot', async ({ app }) => {     // <-- BUG: 'snapshot'
  const svc = app.get(ShuttleTrackerService);
  await svc.fetchRoutesAndStops();
});
```
This can cause two problems:
1. Advisory lock collision: `snapshot` is the lock key for `snapshot.ts` (every 15 min). When midnight hits and `fetch-routes-stops` runs, it holds the `snapshot` lock for the duration of the Passio HTTP roundtrip. The 12:00 AM snapshot cron racing 1 ms later sees the lock held and skips that 15-min snapshot. Silent data loss.
2. Heartbeat misfire: since PR #122 wires per-job heartbeats. With `snapshot` here, a daily routes ping will hit the 15-min snapshot heartbeat URL -> no missed-snapshot alert ever fires + the new `BETTERSTACK_HEARTBEAT_FETCH_ROUTES_STOPS` (when set) gets zero pings -> false-positive page.

Fix: rename `fetch-routes-stops` (and add a `BETTERSTACK_HEARTBEAT_FETCH_ROUTES_STOPS` Better Stack monitor + Fly secret in a follow-up). Ping me if you need help configuring those.

#### 2. `fetchRoutesAndStops` runs at app startup (and never expires) -> cold-start latency + module-init failure mode
In `shuttle-tracker.service.ts`:
```typescript
async onModuleInit() {
  await this.fetchRoutesAndStops();
  await this.fetchShuttles();
}
```
This blocks the Nest bootstrap on two third-party HTTP calls to PassioGO! every time the API container boots. Risks:
- If PassioGO is slow (5-10s), Fly health checks fail -> container marked unhealthy -> cold-start storm.
- If PassioGO is down at boot, the API still boots but with empty `currentRoutes` and `currentStops`. The cron will refresh at midnight, but until then `/transit/routes` returns `[]` with no observability signal.
- This couples API startup health to a third-part that is *not in our SLO*.
Fix: make `onModuleInit` non-blocking and add error-state visibility:
```typescript
async onModuleInit() {
  // Fire-and-forget so module init doesn't block on PassioGO
  void this.fetchRoutesAndStops().catch((err) =>
    this.logger.error('Initial route fetch failed; will retry on next cron tick', err),
  );
}
```
I think we can also drop `fetchShuttles()` from init entirely since the WS gateway will populate live data within seconds anyway.

#### 3. In-memory state on multi-instance deploy -> wrong data per request
`latestShuttles`, `currentRoutes`, `currentStops` live in service instance memory:
```typescript
private latestShuttles: MapShuttle[] = [];
private currentRoutes: MapRoute[] = [];
private currentStops: MapStop[] = [];
```
Today on Fly we run one API container, so this works. But once we scale to 2 (which we'll do for launch, even just for HA), each container has its own copy:
- WebSocket subscribes to PassioGO from each container -> 2x outbound API hits + duplicate broadcasts
- REST `/transit/routes` returns whichever container's snapshot the load balancer picked -> different clients see different data
- Cron `fetch-routes-stops` only refreshes the *cron container's* memory, not API containers'

Fix options (in order of preference):
1. Cache in Redis: we can add Redis to the stack (Fly Redis $0 free tier), use it for shared shuttle state. Both API containers read from it, cron writes to it. Standard answer.
2. Cache in Postgres: `shuttle-state` table with single row + JSONB columns. No new infra, slower hot path.
3. Document "single instance only" + add a Fly `[deploy] strategy = "immediate" constraint`: Only acceptable if you commit to never scaling, I wouldn't advise this for launch.

#### 4. Mobile shuttle URLs are hard-coded to `http://api.sharkpark.app/shuttles` (wrong scheme and path)
```typescript
const getShuttleUrl = (): string => {
  if (!__DEV__) {
    return 'http://api.sharkpark.app/shuttles';   // ← http:// + wrong path
  }
```
Main problems:
1. `http://` not `https://`: App Transport Security on iOS will block the connection. App store rejects too.
2. `/shuttles` path doesn't match the gateway's actual mount (`/api/v1/socket.io/` namespace `shuttles`).
3. Feel free to correct me if I'm wrong, but I'm not sure why this is a separate URL at all, we should use `API_CONFIG.BASE_URL` (or its origin) and let `path: '/api/v1/socket.io/'` do the routing. We can drop `getShuttleUrl()` entirely.

```typescript
// In useTransitData:
const socket = io(new URL(API_CONFIG.BASE_URL).origin + '/shuttles', {
  transports: ['websocket'],
  path: '/api/v1/socket.io/',
});
```

#### 5. All four `transit/*` endpoints are `@Public()` -> bypass the access-tier contract
Per `docs/api-access-tiers.md`, shuttle data is Public layer by intent, but `@Public()` also bypasses the tier-aware throttler that our access implementation introduced.
Main problems:
1. PassioGo is the actual data source and we're effectively reverse-proxying it. With public + no throttle, any scraper can DoS our API to scrap PassioGO via us -> we get rate-limited by PassioGO and our shuttle feature breaks for real users.
2. The WebSocket gateway has `cors.origin: '*'` and no auth at all. Anyone can connect from any browser and consume our PassioGO upstream forever.

Fix:
- ` Keep `@Public()` on the REST endpoints, but apply a tight `@Throttle({ default: { limit: 30, ttl: 60_000 } })` per endpoint or controller-wide. Per-IP is fine here.
- Lock the WS gateway's `cors.origin` to `['https://sharkpark.app', 'http://localhost:8081']` at minimum, and add a JWT verification handshake (or at least a shared signing key check) so random web tabs can't open sockets.

### Some nits:
1. `fetchShuttles()` exists but is never called by a cron or WS, and overlaps with the live feed
- The WS path delivers shuttle updates continuously, but `fetchShuttles()` is only invoked at module init (which #B2 says we should drop). Either delete the REST polling fallback entirely, or document when it's intended to be called (cron? on subscribe? after WS gap?). Right now it's dead code from boot+1s onward.
2. `passio-websocket.service.ts` lacks exponential backoff
- Reconnect is hard-coded 5s. If PassioGO is down for a while, we'll hammer them. Add jitter + cap (`min(5s * 2^n + jitter, 5min)`).
3. `WebSocket` import is the entire `ws` package
- `import WebSocket from 'ws'` brings in 200KB at startup. Fine, just FYI for the bundle-size watcher.
4. `IOS_IP = '192.168.1.113'` hard-coded back into config
- This was supposed to be solved by `SHARKPARK_API_URL` env override (PR #82). Reverting to a hard-coded LAN IP for iOS will hand the next dev a confusing "works on my machine" experience. Pull from `@env` or comment that it must be set explicitly.
5. `console.log` debug spam in `transit.ts`
```typescript
if (__DEV__) {
  console.log('[transitService] Routes Data:', routesResponse.data);
}
```
- Routes payload is huge. Even guarded by `__DEV__` this floods Metro. Use the existing logger or make it opt-in via a flag.
6. `shuttle-tracker.service.ts` has an unfixed TODO in production code
`// TODO: verify that this fetch works with active busses en route`
- Either verify and remove, or open a follow-up issue and replace with `// see issue #NNN`.

### Questions:
1. Is the Google Maps API key in `local.properties` git-ignored?
- The build.gradle wires `MAPS_API_KEY` from `local.properties`. That file is usually .gitignored by RN templates, but please confirm. If it's committed, rotate the key immediately. Also add an `.env.example`-style `local.properties.example` documenting the expected key.
2. What's the Google Maps billing posture?
- Google Maps for Android Native has a free quota but charges past it (~$200/mo of free credit). At launch traffic this is fine, but:
  - Restrict the API key to "Android apps" + your specific package + SHA-1 cert in Google Cloud Console.
  - Set a billing alert for $50/mo so we're not surprised.

zachfpadilla:
Thank you for the thorough review @chuckmilton!

I've implemented all of the suggested changes & have additionally updated the main README with a few of the user-centric shuttle additions.

I've added `fetchShuttles()` to the `fetch-transit` (formerly `fetch-stops-routes`) cron job as was originally intended - its job is to pull static data that isn't & doesn't need to be constantly received via the websocket (e.g. name & color). The websocket is currently configured to *only* send constantly-updated data (e.g. coords).

`cron: fetch-transit` now exists in BetterStack and exists as a secret in Fly as `BETTERSTACK_HEARTBEAT_FETCH_TRANSIT`.

I've also added Redis to our stack. `REDIS_URL` is a new Fly secret that has been configured for a new `sharkpark-cache` container that currently exists to centralize the storage of transit data (routes, stops, & shuttles). As it stands, PassioGO via our websocket is independent of this (though I don't think that should matter, as it's served through a publicly-accessible website anyways).

In terms of the Google Maps API, it's free for the basic map display. The key's already been restricted to our package ID (com.sharkpark.app) & the `/android/app/debug.keystore` SHA-1 fingerprint (I'll reconfigure this once the prod key is generated).

I've also added `/android/local/local.properties.example`, which shows how to place the Maps API key into `local.properties` proper.

Let me know if there's anything else! gassed af rn bro

chuckmilton:
> Thank you for the thorough review @chuckmilton!
> 
> I've implemented all of the suggested changes & have additionally updated the main README with a few of the user-centric shuttle additions.
> 
> I've added `fetchShuttles()` to the `fetch-transit` (formerly `fetch-stops-routes`) cron job as was originally intended - its job is to pull static data that isn't & doesn't need to be constantly received via the websocket (e.g. name & color). The websocket is currently configured to _only_ send constantly-updated data (e.g. coords).
> 
> `cron: fetch-transit` now exists in BetterStack and exists as a secret in Fly as `BETTERSTACK_HEARTBEAT_FETCH_TRANSIT`.
> 
> I've also added Redis to our stack. `REDIS_URL` is a new Fly secret that has been configured for a new `sharkpark-cache` container that currently exists to centralize the storage of transit data (routes, stops, & shuttles). As it stands, PassioGO via our websocket is independent of this (though I don't think that should matter, as it's served through a publicly-accessible website anyways).
> 
> In terms of the Google Maps API, it's free for the basic map display. The key's already been restricted to our package ID (com.sharkpark.app) & the `/android/app/debug.keystore` SHA-1 fingerprint (I'll reconfigure this once the prod key is generated).
> 
> I've also added `/android/local/local.properties.example`, which shows how to place the Maps API key into `local.properties` proper.
> 
> Let me know if there's anything else! gassed af rn bro



> Thank you for the thorough review @chuckmilton!
> 
> I've implemented all of the suggested changes & have additionally updated the main README with a few of the user-centric shuttle additions.
> 
> I've added `fetchShuttles()` to the `fetch-transit` (formerly `fetch-stops-routes`) cron job as was originally intended - its job is to pull static data that isn't & doesn't need to be constantly received via the websocket (e.g. name & color). The websocket is currently configured to _only_ send constantly-updated data (e.g. coords).
> 
> `cron: fetch-transit` now exists in BetterStack and exists as a secret in Fly as `BETTERSTACK_HEARTBEAT_FETCH_TRANSIT`.
> 
> I've also added Redis to our stack. `REDIS_URL` is a new Fly secret that has been configured for a new `sharkpark-cache` container that currently exists to centralize the storage of transit data (routes, stops, & shuttles). As it stands, PassioGO via our websocket is independent of this (though I don't think that should matter, as it's served through a publicly-accessible website anyways).
> 
> In terms of the Google Maps API, it's free for the basic map display. The key's already been restricted to our package ID (com.sharkpark.app) & the `/android/app/debug.keystore` SHA-1 fingerprint (I'll reconfigure this once the prod key is generated).
> 
> I've also added `/android/local/local.properties.example`, which shows how to place the Maps API key into `local.properties` proper.
> 
> Let me know if there's anything else! gassed af rn bro

Awesome work @zachfpadilla! I appreciate you for getting all of those revisions done so quickly. Everything LGTM 💯

I made a few small changes to further refine the revisions. Feel free to check these out whenever:

- **Types** (shuttle-tracker.interface.ts, transit.ts): added `routeId` to `MapStop`/`MapShuttle`, `status`/`name`/`shortName` to `MapRoute`.
- **Backend service** (shuttle-tracker.service.ts): fixed double-slash in PassioGO URL.
- **WS validation** (passio-websocket.service.ts): `handleShuttleUpdate` now async, validates frames via `plainToInstance` + `validateOrReject`; malformed frames dropped with warn log.
- **Gateway hardening** (shuttle-tracker.gateway.ts): production fails closed when `WS_CONNECT_SECRET` is unset.
- **Mobile hook** (useTransitData.ts): inserts placeholder for unknown shuttle IDs and triggers static metadata refresh via `shuttlesRef`.
- Tests updated; full suites green (backend 496/496, mobile 609/609); lint + typecheck pass.

PR #121, opened 2026-04-28
feat: reports endpoint

zachfpadilla:
# POST `/api/v1/reports` Endpoint
## Summary
Implements the `${BASE_URL}/reports` endpoint. This will be used to POST to Prisma anytime a report is submitted by a user via the frontend ReportModal within `ShortTermForecastScreen.tsx`.

### `add_lot_reports` Prisma migration & `/reports` endpoint
- A new Report model & ReportType enum (`BLOCKAGE`, `CRASH`, `OTHER`) have been added to the schema.
- Reports contain the lot cuid (string), type (ReportType), and message (optional string).
### Usage
- A POST request with a body containing:
  - **lotId** (string): id of a prisma Lot model instance (e.g. *NOT* 'G2')
  -  **type** ('blockage' | 'crash' | 'other'): string representation of a ReportType value
  -  **message** (string; optional): additional message (if user selects 'other')
### Example
```typescript
import apiService, { ApiResponse } from './base';

// CreateReportDto payload mirror
export interface CreateReportPayload {
  lotId: string;
  type: 'blockage' | 'crash' | 'other';
  message?: string;
}

// CreateReportDto response shape
export interface CreateReportResponse {
  id: string;
  created_at: string;
}

export const submitIncidentReport = async (
  payload: CreateReportPayload
): Promise<ApiResponse<CreateReportResponse>> => {
  return await apiService.post<CreateReportResponse>('/reports', payload);
};

const reportExample: CreateReportPayload = {
  lotId: 'cm0abc1230000xyz', // id (cuid) from Lot prisma model
  type: 'other',
  message: 'G1 is a myth'
};

const run = async () => {
  try {
    const response = await submitIncidentReport(reportExample);

    if (response.success) {
       console.log('report submitted successfully. yippee!!!');
       console.log(`Tracking ID: ${response.data.id}`);
       console.log(`Timestamp: ${response.data.created_at}`);
    }
  } catch (error) {
    console.error('Failed to submit report! :(((', error);
  }
};

run();
```


chuckmilton (commented):
Solid scaffolding overall. The Prisma model, DTO, module wiring, and e2e shape look pretty good! I do have some suggestions regarding the security/identity model and multi-tenant correctness for this. None of the blocks are big lifts, so they should be quick to fix.

### Requested Changes:

Without this, the throttler will also be useless, it would be IP-keyed across all of CSULB's NAT'd campus wifi.

1. Add `user_id` FK to `Report`
The row as no link to the reporter. We need the FK for per-user throttling, @DottL's reliability "weight 0 for repeat-flagged user" rule, `DELETE /users/me` cascade for App Store / GDPR, admin review queue, future "your reports" history in Settings.

Privacy posture: stored internally, never exposed via any API. Disclose in privacy nutrition label as "Identifies - Linked to identity, used for app functionality." Same pattern as Reddit or Discord for example. Fully App Store/GDPR compliant.

```
model Report {
  id         String     @id @default(cuid())
  lot_id     String
  user_id    String     // NEW
  type       ReportType
  message    String?
  created_at DateTime   @default(now())

  lot  Lot  @relation(fields: [lot_id], references: [id], onDelete: Cascade)
  user User @relation(fields: [user_id], references: [id], onDelete: Cascade)

  @@index([lot_id, created_at], name: "idx_reports_lot_time")
  @@index([user_id, created_at], name: "idx_reports_user_time")
  @@map("reports")
}
// + add `reports Report[]` to User model
```
Controller pulls `user_id` from `req.user.id` (AzureAdGuard attached it). Never accept from request body.

2. Lot lookup not school-scoped
```typescript
const lot = await this.prisma.lot.findFirst({
  where: { lot_id: dto.lotId },
});
```

`lots` is `@@unique([school_id, lot_id])`,  not unique by `lot_id` alone. Today CSULB is the only school so this works, but once we onboard a 2nd school, lots will collide and reports will silently attach to the wrong school's lot.

Fix: mobile passes to cuid (`lot.id`) directly. You can coordinate with @lawrenceDegoma  on wire format.

### Some nits:
1. No throttler override
Since this is now per-user once we add getTracker() (separate ticket), tighter limits make sense; for now it's IP-keyed and shared with the rest of the app.
```typescript
@Throttle({ default: { limit: 5, ttl: 60_000 } })
```

2. Service returns row, controller discards it
We should either return the created report (clients can use `id` for tracking/retry/Sentry correlation) or have the service return `Promise<void>`. I recommend returning `{id, created_at}`.

3. `typeMapping` table is unnecessary
DTO and Prisma enums are isomorphic:
```typescript
type: dto.type.toUpperCase() as ReportType,
```

4. `e2e exceptionFactory` duplicates `main.ts`
The custom error formatter in `reports.e2e-spec.ts` re-implements `ValidationPipe` config. If `main.ts` ever changes, the test won't catch the regression. If other e2e suites have a `bootstrapTestApp()` helper, use it. Otherwise, file a follow-up.

### Lint state
Lint: 0 errors, 1 warning (unused eslint-disable in `reports.controller.spec.ts`, drop the directive)

zachfpadilla:
Thank you for the review @chuckmilton!

I've since made the suggested changes & have defined the endpoint return shape as `{id, created_at}`. Per your last nit, I also created `/backend/tests/utils/bootstrap.ts` to mirror the config in `main.ts` that's up to this point been copy/pasted between all of our e2e test suites.

chuckmilton (approved):

PR #122, opened 2026-04-28
feat(backend): Better Stack heartbeat pings on cron success

chuckmilton:
## Summary

Wire each cron script to ping a Better Stack heartbeat URL on success so we get alerted within minutes when a cron silently stops running (which today has zero monitoring beyond the Sentry alert rules from P14.125).

## Changes

- **`apps/backend/src/scripts/_heartbeat.ts`** — new `pingHeartbeat(jobName, logger?)` helper. Resolves URL from env var `BETTERSTACK_HEARTBEAT_<JOB_NAME_UPPER_SNAKE>`, GETs with a 5s AbortController timeout, swallows all errors, logs warn on failure.
- **`apps/backend/src/scripts/_bootstrap.ts`** — `runCronJob` now pings the heartbeat after success or busy-lock skip (the lock holder will ping when it finishes; pinging on the skipped instance prevents a false-positive missed-heartbeat alert). Hard failure path skips the ping intentionally.
- **`apps/backend/src/scripts/_heartbeat.spec.ts`** — 6 unit tests covering the env-key derivation, no-op when unset, GET success, non-2xx warning, network-error warning, and logger-optional usage.
- **`apps/backend/eslint.config.mjs`** — register Node 18+ web globals (`fetch`, `AbortController`, `Response`, `globalThis`) so we don't need a per-file `/* eslint-disable no-undef */`.
- **`apps/backend/src/weather/weather-fetch.service.ts`** — drop the now-redundant eslint-disable directive.

## Crons covered (6)

| Cron | Schedule | Better Stack period | Grace |
|---|---|---|---|
| `snapshot` | every 15 min | 15 min | 5 min |
| `fetch-weather` | every 30 min | 30 min | 10 min |
| `cleanup-device-states` | daily 3am PT | 1 day | 30 min |
| `backup-db` | daily 2am PT | 1 day | 30 min |
| `verify-latest-backup` | weekly Mon 4am PT | 7 days | 60 min |
| `prune-old-data` | daily 4am PT (skip Mon) | 1 day | 30 min |

## Post-merge step (Charles)

Set the 6 Fly secrets (URLs already created in Better Stack):

```bash
fly secrets set \
  BETTERSTACK_HEARTBEAT_SNAPSHOT='https://uptime.betterstack.com/api/v1/heartbeat/...' \
  BETTERSTACK_HEARTBEAT_FETCH_WEATHER='...' \
  BETTERSTACK_HEARTBEAT_CLEANUP_DEVICE_STATES='...' \
  BETTERSTACK_HEARTBEAT_BACKUP_DB='...' \
  BETTERSTACK_HEARTBEAT_VERIFY_LATEST_BACKUP='...' \
  BETTERSTACK_HEARTBEAT_PRUNE_OLD_DATA='...' \
  -a sharkpark-api
```

Setting secrets triggers a redeploy. First snapshot heartbeat should arrive in Better Stack within 15 min of deploy completing.

## Test plan

- [x] `pnpm --filter backend test` — 13 tests pass (6 new + 7 existing)
- [x] `pnpm --filter backend lint` — 0 errors
- [x] `pnpm --filter backend build` — clean
- [ ] Post-merge: confirm all 6 monitors flip green in Better Stack within their first cron window

## Behavior notes

- **Lock-skip case pings:** if two cron containers race, one acquires the advisory lock and the other no-ops. Both ping. This is intentional — a silent skip without a ping would alert as 'missed heartbeat' even though the system is healthy.
- **Hard failure case does NOT ping:** an exception during `work` skips the ping. That's what triggers the 'missed heartbeat' alert and surfaces the bug. Sentry catches the error in parallel via the existing instrument.ts hook.
- **Local dev / tests are inert:** no env var → no ping → no accidental production-monitor noise.


zachfpadilla (approved):
Nice! I was actually planning on asking you how we were planning to monitor crons. Anything that makes the linter kinder is also appreciated. LGTM

chuckmilton:
> Nice! I was actually planning on asking you how we were planning to monitor crons. Anything that makes the linter kinder is also appreciated. LGTM

Thank you for the kind review @zachfpadilla (Zachariah)!

PR #139, opened 2026-04-30
feat(mobile): wire 403 BG_LOCATION_REQUIRED → two-stage location perm screen

lawrenceDegoma:
### What
When the backend's `ContributorGuard` rejects a request with `403 { code: 'BG_LOCATION_REQUIRED' }` (device hasn't sent a ContributorPing within the 30-min TTL), the app now:
1. Catches the error at the API layer
2. Surfaces a `bgLocationRequired` flag up through the data hooks
3. Navigates from `MapScreen` to a new `LocationPermissionScreen` that walks the user through enabling background location

### Why
App Store guideline 5.5.4 (and our own UX contract) requires that we explain why Always-on location is needed before invoking the system prompt. This screen satisfies that requirement and makes the contributor re-onboarding flow explicit rather than silently failing.

### Changes

| File | Change |
|------|--------|
| `services/api/base.ts` | New `BgLocationRequiredError` class; 403 body parsed in `attemptRequest` |
| `services/api/index.ts` | Re-exports `BgLocationRequiredError` |
| `types/navigation.ts` | `LocationPermission: { reason?: string }` added to `MapStackParamList` |
| `screens/LocationPermissionScreen.tsx` | **New** — four-step state machine: `explain → when-in-use → always (opt-in) → done` |
| `screens/index.ts` | Barrel export for new screen |
| `navigation/MainTabNavigator.tsx` | `LocationPermission` screen added to `MapStack` |
| `hooks/useLotData.ts` | `bgLocationRequired` flag exposed from both `useLotData` and `useLotsList` |
| `screens/MapScreen.tsx` | `useEffect` navigates to `LocationPermission` when flag is true |
| `types/images.d.ts` | Wildcard module declarations for `.webp/.png/.jpg/.jpeg` (fixes pre-existing tsc error) |
| `tsconfig.json` | Override `allowArbitraryExtensions: false` so wildcard `declare module '*.webp'` resolves |

### User flow
API call fails with 403 BG_LOCATION_REQUIRED → useLotsList sets bgLocationRequired = true → MapScreen useEffect → navigate('LocationPermission') → Step 1 (explain): what's collected / what isn't → "Enable location access" → Step 2 (WhenInUse system prompt) → Granted → Step 3 (Always opt-in, soft ask only) → "Allow background" or "No thanks" → Step 4 (done)
User can press the back arrow at any step — they are **never hard-blocked**.

### Testing
- `pnpm --filter mobile exec tsc --noEmit` → 0 errors


chuckmilton:
Hey @lawrenceDegoma — pushed `4402f13` to your branch with the merge from `main` and the review fixes baked in. Wanted to walk you through what changed and why so nothing's a surprise:

## What I did

**Merged `main` (not rebased)** so all your commits stay intact in the history. No force-push.

### Conflicts (all 6 from #120 Dynamic Map landing on main)

The big one: **`main` already had a `BackgroundLocationRequiredError`** with a richer design than your `BgLocationRequiredError`:
- Extends `ApiError` (not raw `Error`), so existing `instanceof ApiError` checks in retry logic still work
- Has a static `CODE = 'BG_LOCATION_REQUIRED'` constant
- The `isRetryable()` helper already special-cases it to suppress retries on the contributor gate
- A dedicated `parseBgLocationRequired()` helper handles the JSON parse safely

Rather than have two parallel error classes, I dropped yours and wired everything to `BackgroundLocationRequiredError`. Your hook & test were renamed to match.

The other conflicts were mechanical (combine your `LocationPermissionScreen` import with main's `CustomText` import, keep both screen exports, take main's superset `images.d.ts`).

### Headline blocker fix — the `ready()` collision

`LocationPermissionScreen.requestWhenInUse()` was calling `BackgroundGeolocation.ready({...})` with a WhenInUse override, but the **Transistor SDK only honors the first `ready()` call per app lifecycle**. The production config (Always auth, `PersistMode.None`, headless mode, polygon geofences) is already applied at app boot in `locationService.initialize()` — your override silently lost.

Both `requestWhenInUse()` and `requestAlways()` now route through the existing `locationService.requestPermissions()` which:
- Wraps `BackgroundGeolocation.requestPermission()`
- **Also** handles iOS 14+ `requestTemporaryFullAccuracy('ParkingDetection')` when the user grants Reduced accuracy (we'd have lost this otherwise)

### Other fixes from review

- **Nav re-trigger loop**: added `clearBgLocationRequired()` to both hook returns; `MapScreen` now guards the navigate effect on `useIsFocused()` and clears the flag before pushing — so back-button doesn't bounce them right back in.
- **Dropped `tsconfig.json` `allowArbitraryExtensions: false`** — it's the TS 5.0+ default and doesn't enable wildcard module declarations the way the change implied.
- Removed unused `BackgroundGeolocation` + `Platform` imports from the screen.
- Added `__tests__/LocationPermissionScreen.test.tsx` smoke test (explain → always on grant; explain → done on deny).

## Verified

- `pnpm --filter mobile lint` ✅
- `pnpm --filter mobile exec tsc --noEmit` ✅
- `pnpm exec jest` → **612/612 passing**

Take a look at the diff when you get a chance and let me know if anything reads off — happy to iterate. Solid first pass on the contract; the SDK gotcha was subtle.

PR #146, opened 2026-05-01
Add user reports factor to reliability score

DottL:
# Wire user reports into reliability scoring

Adds user reports as a sixth factor in the per-lot reliability score. When users flag a lot as inaccurate, its confidence badge drops from "Reliable" toward "Moderate" or "Estimate", and the breakdown modal calls out recent reports as the reason. Distinct-reporter dedup at the query layer prevents a single user from swinging the score by spamming the report button.

### Backend — new factor `userReports` in the existing weighted formula

- **Sixth factor**, weighted at **0.15**. The existing five sensor factors were shrunk by 15% so weights still sum to 1.0:

  | Factor | Before | After |
  |---|---|---|
  | `penetrationRate` | 0.35 | 0.30 |
  | `dataFreshness` | 0.25 | 0.21 |
  | `eventFrequency` | 0.20 | 0.17 |
  | `sampleSize` | 0.15 | 0.13 |
  | `historicalAccuracy` | 0.05 | 0.04 |
  | **`userReports`** | — | **0.15** |

- **Framing:** factor reads as "absence of reports" so high = good, matching every other factor's semantics. 
  - `normalizedValue = 1 - min(1, reporterCount / target)` with `target = 5`. 
  - No reports -> 1.0 (full credit)
   - 5+ distinct reporters in the window -> 0.0 (no credit)
- **Per-(user, lot) dedup at the query layer** via `prisma.report.findMany({ distinct: ['user_id'] })`. A single user mashing the report button counts as 1, not N. Closes the most common spam vector without admin tooling or schema changes.
- **Window:** 60 minutes by default, read from `thresholds.userReportsWindowMinutes` so it tracks the same config path as the other thresholds.

### Mobile — modal automatically shows the new factor

- Added `userReports` to `ReliabilityFactors` and `ReliabilityConfig` types so the API response shape matches.
- Added `userReports: 'User Reports'` to `formatFactorName` in `ReliabilityModal` so the breakdown row reads "User Reports" instead of the raw key.
- No component changes — the existing per-factor bar rendering picks up the new factor automatically.


apps/backend/src/reliability/reliability-computation.service.ts
chuckmilton, line 33: **N+1 query.** `computeReliabilityForAllLots` already loops `gatherReliabilityInput` per lot, and this PR adds another per-lot round-trip. With 28 lots that's 28 extra DB calls per `/reliability/lots` request.

Suggested follow-up (doesn't have to land in this PR, but worth filing):

```ts
const reporterCounts = await this.prisma.report.groupBy({
  by: ['lot_id', 'user_id'],
  where: { created_at: { gte: reportsWindowStart, lte: now } },
});
const distinctByLot = new Map<string, number>();
for (const row of reporterCounts) {
  distinctByLot.set(row.lot_id, (distinctByLot.get(row.lot_id) ?? 0) + 1);
}
```
Then pass `distinctByLot.get(lot.id) ?? 0` into each per-lot `gatherReliabilityInput`.

apps/backend/src/reliability/reliability-computation.service.ts
```diff
@@ -106,12 +106,25 @@ export class ReliabilityComputationService {
     // Get historical accuracy if available
     const historicalAccuracy = await this.getHistoricalAccuracy(lotId);
 
+    // Distinct reporters in the configured window - deduped per user so spamming counts as 1.
+    const { userReportsWindowMinutes } = this.reliabilityService.getDefaultThresholds();
```
chuckmilton: Minor coupling concern: this reads from `getDefaultThresholds()` directly, which means a caller that passes custom thresholds to `computeReliability` would compute the report window from the defaults but score it against custom thresholds.

Not a problem today (no caller does that), but if/when we make thresholds per-lot or A/B-testable, this will silently desync. Worth threading thresholds through `gatherReliabilityInput` as a parameter.

apps/backend/src/reliability/reliability.service.ts
```diff
@@ -173,12 +200,25 @@ export class ReliabilityService {
...
       factors.sampleSize,
+      factors.userReports,
     ];
 
     const weakestFactor = factorArray.reduce((min, f) =>
       f.normalizedValue < min.normalizedValue ? f : min,
     );
 
+
+    if (weakestFactor === factors.userReports) {
```
chuckmilton: Tiny nit: the `case 'HIGH':` branch here returns the exact same string as the default `case 'HIGH':` below. The whole `if` could fall through for HIGH and only special-case MEDIUM/LOW:

```ts
if (weakestFactor === factors.userReports && confidence !== 'HIGH') {
  return confidence === 'MEDIUM'
    ? 'Moderate confidence. Recent user reports suggest the live count may not match conditions on the ground.'
    : 'Low confidence due to recent user reports. Use estimates with caution.';
}
```
Not blocking.

apps/backend/src/reliability/reliability.service.ts
```diff
@@ -115,6 +119,29 @@ export class ReliabilityService {
...
       ),
+      userReports: this.computeUserReportsFactor(
+        input.uniqueReportersInWindow ?? 0,
+        weights.userReports,
+        thresholds.userReportsTarget,
+      ),
+    };
+  }
+
+  // High = good (absence of reports). Reporter count is deduped per user at the query layer.
```
chuckmilton, line 130: Two thoughts on the model itself. Both probably belong in a follow-up issue rather than this PR, but flagging:

1. **No time decay.** A 5-reporter spike at minute 0 keeps the factor pinned at 0 for the full 60-minute window, then jumps to 1 the moment the oldest report ages out. A weighted-by-recency variant would feel less binary:
```ts
   const penalty = reports.reduce((sum, r) => {
     const ageMs = now - r.created_at.getTime();
     const weight = Math.max(0, 1 - ageMs / windowMs);
     return sum + weight;
   }, 0) / safeTarget;
```

2. **Spam vector**. Distinct-by-`user_id` is a good first pass, but 5 distinct accounts = full penalty. Worth considering: gate reports on the user being inside the geofence (we already collect location), require the user to be a contributor, or scale `target` by lot capacity so a 1000-stall lot needs more reporters than a 50-stall lot.

apps/backend/src/reliability/reliability-computation.service.spec.ts
```diff
@@ -121,6 +124,45 @@ describe('ReliabilityComputationService', () => {
...
+
+      const call = prisma.report.findMany.mock.calls[0][0];
+      const gte: Date = call.where.created_at.gte;
+      const lte: Date = call.where.created_at.lte;
+      const windowMs = lte.getTime() - gte.getTime();
+      
+      expect(windowMs).toBe(30 * 60 * 1000);
+      expect(lte.getTime()).toBeGreaterThanOrEqual(before);
+      expect(lte.getTime()).toBeLessThanOrEqual(after);
+    });
```
chuckmilton, line 179: The window-width assertion is good, but the boundary check (`lte.getTime() >= before && <= after`) only catches gross drift. If you want it tighter, capture `now` once before the await and assert `gte.getTime() === now.getTime() - 30 * 60 * 1000` exactly. Not blocking.

chuckmilton (commented):
Great work on the user reports factor for our reliability score! The new factor slots into the existing weighted-scoring framework really cleanly. Weights still sum to exactly 1.0, the "absence of reports = good" framing keeps semantics consistent with the other five factors, and doing the per-user dedup at the query layer (instead of in the service) is exactly the right call. Test coverage is thorough too, full-credit, mid-target, clamp-to-zero, the HIGH→MEDIUM movement test, and the report-aware explanation copy are all covered.

I left a few inline comments, but nothing blocking, mostly follow-up ideas around an N+1 in `computeReliabilityForAllLots`, the spam surface (5 distinct accounts still zeros the factor), and the hard-cutoff window with no decay. Other than that, LGTM 💯

apps/backend/src/reliability/reliability-computation.service.spec.ts
DottL: Kept range check over exact now - 30min equality — service calls new Date() internally, so exact match would need fake timers. 😿

DottL:
Thank you for the feedback @chuckmilton🎉. I've since made some changes for you to take a look at one last time.

----
## Update
- **Batched the all-lots reliability query.** `/reliability/lots` went from ~112 round-trips (28 lots × 4 queries each, sequential) to 5 round-trips (4 batched + 1 lot fetch), grouped in memory.
- **Shared assembler.** Extracted `assembleReliabilityInput` so the single-lot and batch paths build their `ReliabilityInput` from the same code — drift between the two endpoints is now structurally impossible.
- **Single-lot path also faster.** Its four queries now run in parallel via `Promise.all` instead of sequentially.

- **Recency-weighted MAPE.** `historicalAccuracy` now linearly decays older predictions (weight 1.0 today → 0.1 a week ago), so a recent regression isn't masked by clean older data.
- **Binary search for snapshot matching.** Replaces the per-prediction linear scan, this lets us drop the `take: 100` cap for a more stable MAPE denominator.
- **Misc cleanup.** Threaded `ReliabilityThresholds` through the single-lot path, simplified the user-reports explanation branch, restored per-lot error handling in the all-lots loop.

chuckmilton:
> Thank you for the feedback @chuckmilton🎉. I've since made some changes for you to take a look at one last time.
> 
> ## Update
> * **Batched the all-lots reliability query.** `/reliability/lots` went from ~112 round-trips (28 lots × 4 queries each, sequential) to 5 round-trips (4 batched + 1 lot fetch), grouped in memory.
> * **Shared assembler.** Extracted `assembleReliabilityInput` so the single-lot and batch paths build their `ReliabilityInput` from the same code — drift between the two endpoints is now structurally impossible.
> * **Single-lot path also faster.** Its four queries now run in parallel via `Promise.all` instead of sequentially.
> * **Recency-weighted MAPE.** `historicalAccuracy` now linearly decays older predictions (weight 1.0 today → 0.1 a week ago), so a recent regression isn't masked by clean older data.
> * **Binary search for snapshot matching.** Replaces the per-prediction linear scan, this lets us drop the `take: 100` cap for a more stable MAPE denominator.
> * **Misc cleanup.** Threaded `ReliabilityThresholds` through the single-lot path, simplified the user-reports explanation branch, restored per-lot error handling in the all-lots loop.

Thank you for getting the revisions done! I pulled in your changes and everything is working as expected. Great work! 🤓

chuckmilton (approved):

PR #147, opened 2026-05-01
feat(backend): push notifications setup

zachfpadilla:
# Push Notification Infrastructure
## Summary
Adds the full backend sender infrastructure for push notifications. Checks every 15 minutes for whether or not to send any of the 4 types of notifications that users can opt-in for. Also logs already-sent notifications so that a given user doesn't receive the same push notification twice.

### Schema & migration
- **`push_tokens`** — stores FCM device tokens per user (`user_id`, `token` unique, `platform`, `created_at`). Upserts on `token` so reinstalls and token refreshes don't create duplicates.
- **`notification_logs`** — dedup log that prevents sending the same alert class to the same user twice within the configured window. `lot_id` is overloaded to hold `event_id` for the events trigger.

### New endpoint (`POST /users/me/push-token`)
- Takes in a valid token (via `@react-native-firebase/messaging`) & a platform ('ios' or 'android'). Logs a `push_token` for the user.
- Example (initialization):
```typescript
import messaging from '@react-native-firebase/messaging';
import { registerPushToken } from './services/api/notifications';

async function setupPushNotifications() {
  // iOS: request permission first
  const authStatus = await messaging().requestPermission();
  const enabled =
    authStatus === messaging.AuthorizationStatus.AUTHORIZED ||
    authStatus === messaging.AuthorizationStatus.PROVISIONAL;

  if (!enabled) return;

  const token = await messaging().getToken();
  await registerPushToken(token, Platform.OS as 'ios' | 'android');

  // Re-register if FCM rotates the token
  return messaging().onTokenRefresh(async (newToken) => {
    await registerPushToken(newToken, Platform.OS as 'ios' | 'android');
  });
}
```
- Example (usage):
```typescript
import { apiService } from './base';
import { loadAuth } from '../../auth';

export async function registerPushToken(token: string, platform: 'ios' | 'android'): Promise<void> {
  const auth = await loadAuth();
  if (!auth?.idToken) return;

  await apiService.post<void>('/users/me/push-token', { token, platform }, {
    headers: { Authorization: `Bearer ${auth.idToken}` },
  });

  // { after token logged, update preferences via PATCH /users/:userId/notifications }
}
```

### New cron scripts (`apps/backend/src/scripts/`)
Each runs every 15 minutes, acquires an advisory lock, and exits cleanly if Firebase is not configured. Also added to `/backend/package.json`.

| Script | Trigger | Dedup window |
|---|---|---|
| `notify-favorites-filling.ts` | Favorited lot crosses 80% occupancy | 4h per user + lot |
| `notify-favorites-clearing.ts` | Favorited lot drops below 30% after being above 75% in the last 2h | 4h per user + lot |
| `notify-surge.ts` | Any lot exceeds 90% occupancy (campus-wide) | 2h per user |
| `notify-events.ts` | `CampusEvent.start_time` within the next 2h | 3h per user + event |

### Config & environment
- `FIREBASE_PROJECT_ID`, `FIREBASE_CLIENT_EMAIL`, `FIREBASE_PRIVATE_KEY` added to `.env.example` (must be set for notifications to function).
- `firebase-admin ^12.0.0` added to `package.json`.

## Future Work
- `@react-native-firebase/messaging` integration to obtain FCM tokens (android) / wrap APNs (iOS). See above example.
- As we will now be using 2 Google Cloud services (Google Maps API & Firebase), it may be worthwhile to eventually create a SharkPark [organization](https://docs.cloud.google.com/identity/docs/how-to/set-up-cloud-identity-admin) using the `https://sharkpark.app` domain we own.
   - Both services currently reside within the `SharkPark` project, which in turn resides under `zachfpadilla-org`.


apps/backend/src/notifications/notifications.service.ts
```diff
@@ -0,0 +1,104 @@
...
+    const result = await admin.messaging().sendEachForMulticast({
+      tokens,
+      notification: { title: payload.title, body: payload.body },
+      ...(payload.data ? { data: payload.data } : {}),
+    });
+
+    const stale = result.responses.flatMap((r, i) => (r.success ? [] : [tokens[i]]));
+    if (stale.length > 0) {
+      await this.prisma.pushToken.deleteMany({ where: { token: { in: stale } } });
+    }
```
chuckmilton: **Issue**: This deletes a token on any non-success response from FCM, including transient errors (`messaging/internal-error`, `messaging/server-unavailable`, quota-exceeded, network blips). FCM only treats `messaging/registration-token-not-registered` and `messaging/invalid-registration-token` as "permanently dead" tokens. A single FCM hiccup will nuke valid tokens and silently disable notifications for affected users until they reopen the app.

```suggestion
    const STALE_CODES = new Set([
      'messaging/registration-token-not-registered',
      'messaging/invalid-registration-token',
      'messaging/invalid-argument', // when triggered by token shape
    ]);
    const stale = result.responses.flatMap((r, i) =>
      !r.success && r.error && STALE_CODES.has(r.error.code) ? [tokens[i]] : []
    );
```

Also worth logging non-stale failures via `this.logger.warn` so we don't lose visibility into real send failures.

apps/backend/src/scripts/notify-surge.ts
```diff
@@ -0,0 +1,51 @@
...
+
+void runCronJob('notify-surge', async ({ app, prisma, logger }) => {
+  const svc = app.get(NotificationsService);
+  const since = new Date(Date.now() - SNAPSHOT_WINDOW_MS);
+
+  // Any lot above 90% constitutes a surge — take one to identify the school
+  const surgeSnapshot = await prisma.occupancySnapshot.findFirst({
+    where: { occupancy_rate: { gt: 0.9 }, timestamp: { gte: since } },
+    select: { lot: { select: { school_id: true } } },
+  });
```
chuckmilton: **Issue**: `findFirst` returns one snapshot -> only ONE school gets surge alerts per cron tick. With CSULB-only today this works, but the moment a second school is onboarded this becomes a silent bug where some users stop getting surge alerts depending on Postgres row ordering. Given the rest of the file is school-aware, this seems like a future-you trap.

Fix: group by school:
```suggestion
  const surgingSchools = await prisma.occupancySnapshot.findMany({
    where: { occupancy_rate: { gt: 0.9 }, timestamp: { gte: since } },
    select: { lot: { select: { school_id: true } } },
    distinct: ['lot_id'], // any one snapshot per lot is enough
  });
  const schoolIds = [...new Set(surgingSchools.map((s) => s.lot.school_id))];
```

Then loop over `schoolIds`.

apps/backend/src/scripts/notify-events.ts
```diff
@@ -0,0 +1,59 @@
...
+        notification_preferences: { path: ['event_alerts'], equals: true },
+      },
+      select: { id: true },
+    });
+
+    const startTime = event.start_time.toLocaleTimeString('en-US', {
+      hour: '2-digit',
+      minute: '2-digit',
+      timeZone: 'America/Los_Angeles',
+    });
```
chuckmilton: **Issue**: Timezone is hardcoded to `'America/Los_Angeles'`. The `School` model already has a `timezone` field (defaulting to LA but explicit per-school). When we add a non-PT school, every event push from there will display the wrong start time.

**Fix**: include `school: { select: { timezone: true } }` in the event query and use `event.school.timezone`. Bonus: pulls the format string into a single `formatLocalTime(date, tz)` helper.

apps/backend/src/notifications/notifications.service.ts
```diff
@@ -0,0 +1,104 @@
...
+  private readonly logger = new Logger(NotificationsService.name);
+
+  constructor(private readonly prisma: PrismaService) {}
+
+  onModuleInit() {
+    if (admin.apps.length > 0) return;
+
+    const projectId = process.env.FIREBASE_PROJECT_ID;
+    const clientEmail = process.env.FIREBASE_CLIENT_EMAIL;
+    const privateKey = process.env.FIREBASE_PRIVATE_KEY?.replace(/\\n/g, '\n');
```
chuckmilton: **Issue**: Reads `process.env` directly, even though the PR registers a `notificationsConfig` namespace. Every other module in this codebase consumes config via `ConfigService` / `@Inject(notificationsConfig.KEY)`. Inconsistency aside, direct `process.env` reads can't be overridden in tests without mutating `process.env` globally.

**Fix**: inject `ConfigService` (or the typed config) and read from there.

apps/backend/prisma/schema.prisma
```diff
@@ -265,6 +267,37 @@ model PredictionLongTerm {
...
+}
+
+// Dedup log — prevents sending the same alert class to the same user twice.
+// For lot-level triggers (favorites_filling, favorites_clearing), lot_id is
+// set to the Lot id. For event triggers, lot_id holds the CampusEvent id.
+model NotificationLog {
+  id      String   @id @default(cuid())
+  user_id String
+  type    String   // "favorites_filling" | "favorites_clearing" | "surge" | "events"
+  lot_id  String?
```
chuckmilton: 1. `type` is a free `String`. A typo in any cron script (`'favorite_filling'` vs `'favorites_filling'`) silently breaks dedup, since `hasRecentLog` will look for the typo'd key and find nothing. Promote to a Prisma enum (`NotificationType`) and have the scripts import the enum members.

2. `lot_id` is overloaded to hold `event.id` for the events cron (per the comment in `notify-events.ts`). It also has no FK constraint, which is convenient but smelly. Recommend renaming to `context_id` (no FK), or splitting into `lot_id` + `event_id` columns each with a real FK and `onDelete: SetNull`. The current design will quietly retain log rows referencing deleted events.

apps/backend/src/notifications/notifications.service.spec.ts
chuckmilton, line 1: `notifications.service.spec.ts` is thorough (which is great), but `notify-favorites-filling.ts`, `notify-favorites-clearing.ts`, `notify-surge.ts`, `notify-events.ts` have zero coverage. These contain the actual product logic (thresholds, dedup windows, the "was-high-then-low" check). Given how easy it is to regress the surge-snapshot bug, the stale-token logic, or the JSON-path filters when refactoring, even one happy-path + one dedup-skip test per script would pay off.

apps/backend/src/scripts/notify-favorites-clearing.ts
```diff
@@ -0,0 +1,73 @@
...
+  // Among those, keep only lots that were above 75% in the last 2 hours
+  const clearingLots: { lot_id: string; display_name: string }[] = [];
+  for (const { lot_id, lot } of lowNow) {
+    const wasHigh = await prisma.occupancySnapshot.count({
+      where: {
+        lot_id,
+        occupancy_rate: { gt: 0.75 },
+        timestamp: { gte: historyCutoff, lt: recentCutoff },
+      },
+    });
```
chuckmilton: nit: This fires whenever a lot was ever above 75% in the prior 2h, even if it dropped naturally over an hour. That's probably fine, but if you want to truly capture "lot just cleared," consider checking the most recent prior snapshot rather than `count > 0`, e.g. require the immediately-prior 20-min window's max rate to be > 0.75. Not a blocker.

apps/backend/src/scripts/notify-favorites-filling.ts
```diff
@@ -0,0 +1,54 @@
...
+  for (const { lot_id, lot } of highLots) {
+    const favorites = await prisma.userFavorite.findMany({
+      where: {
+        lot_id,
+        user: { notification_preferences: { path: ['favorites_filling'], equals: true } },
+      },
+      select: { user_id: true },
+    });
+
+    for (const { user_id } of favorites) {
```
chuckmilton, line 42: nit - N+1 in cron loops: Every cron walks lots/events → users → `hasRecentLog` (per-user query) → `sendPush` (per-user query) → `logNotification`. Fine at current user counts; consider batching `hasRecentLog` (`groupBy` on `user_id`) once you cross a few thousand favorites per lot.

This is also referenced in the same pattern in `notify-favorites-clearing.ts` and `notify-events.ts`.

apps/backend/src/notifications/notifications.service.ts
```diff
@@ -0,0 +1,104 @@
...
+        user_id: userId,
+        type,
+        ...(contextId ? { lot_id: contextId } : {}),
+        sent_at: { gte: since },
+      },
+    });
+    return count > 0;
+  }
+
+  async logNotification(userId: string, type: string, contextId?: string): Promise<void> {
```
chuckmilton: nit - TOCTOU between `hasRecentLog` and `logNotification`: Two cron processes (or a retry mid-run) could both pass the dedup check and double-send. Acceptable given the 15-min cron cadence and advisory lock in `runCronJob`, but worth a one-line comment in the service.

apps/backend/prisma/schema.prisma
chuckmilton, line 90: nit - `notification_preferences` defaults all false: The `User` JSON default has all four prefs as `false`, which means existing users get zero notifications until they opt in. That may be intentional (which is good) but worth confirming the mobile side has UI to toggle them, otherwise the entire system is dark by default.

apps/backend/src/notifications/notifications.controller.ts
```diff
@@ -0,0 +1,38 @@
...
+   */
+  @Post('me/push-token')
+  @HttpCode(HttpStatus.NO_CONTENT)
+  async registerPushToken(
+    @Req() req: AuthenticatedRequest,
+    @Body() dto: RegisterPushTokenDto,
+  ): Promise<void> {
+    const email = req.user?.email;
+    if (!email) {
+      throw new ForbiddenException('Authenticated user email missing');
```
chuckmilton: nit - `NotificationsController.registerPushToken` uses `ForbiddenException` for missing email:
Should arguably be `UnauthorizedException`. Missing email on the request user isn't a permission issue, it's an auth issue.

chuckmilton (changes requested):
Great work on the push notification infrastructure, @zachfpadilla (epical)! The structure here are really solid. Schema, service, cron monitor wiring, and the service-level test coverage are all clean and the FCM + iOS SPM setup look right.

I left a handful of inline comments. The two I'd really like addressed before this lands in front of users:

1. Stale-token cleanup is too aggressive: `sendPush` currently deletes a token on any non-success FCM response, which means a transient FCM error or quota blip will nuke valid tokens and silently disable notifications for those users. Should filter to just `messaging/registration-token-not-registered` (and the invalid-token variants).
2. notify-surge only notifies one school per run: `findFirst` picks a single surging snapshot, so once we onboard a second school, surge alerts at one of them will randomly drop depending on row order. Easy fix: `findMany` + dedup school IDs.

The rest is non-blocking polish:
`ConfigService` consistency, hardcoded `'America/Los_Angeles'` in `notify-events` (the `School` model already has a `timezone` field, FYI), promoting `NotificationLog.type` to an enum, and the slightly smelly `lot_id`-as-`event_id` overload. Also, none of the 4 new cron scripts have tests. The service spec is great, but the actual product logic (thresholds, the was-high-then-low check, dedup keys) lives in the scripts and would be cheap to cover.

Lemme know if you have any questions on these suggestions. Happy coding! 🙏

zachfpadilla:
Thank you for the thorough suggestions @chuckmilton (  )!

I've since implemented all of your recommended changes, including the updated prisma migration & 4 new cron test suites. Happy reviewing! 🙏

chuckmilton:
> Thank you for the thorough suggestions @chuckmilton (  )!
> 
> I've since implemented all of your recommended changes, including the updated prisma migration & 4 new cron test suites. Happy reviewing! 🙏

Thank you for making the following revisions! p.s. hey thats me!

PR #148, opened 2026-05-02
fix(mobile): persist isGuest across cold restarts via AsyncStorage

lawrenceDegoma:
## Summary
Fixes `isGuest` being reset to `false` on every cold restart. Previously the state lived only in `useState`, so killing and relaunching the app would always land the user back on the auth screen even if they had chosen "Continue as Guest."

## Changes
- Added `@SharkPark:isGuest` AsyncStorage key (mirrors the `@SharkPark:onboardingComplete` pattern in `useOnboarding.ts`)
- **`initAuth`** — hydrates `isGuest` on mount via `Promise.all([loadAuth(), AsyncStorage.getItem(...)])`, so guest state is restored in the same pass as token loading
- **`continueAsGuest`** — now `async`; persists the flag before setting state
- **`exitGuestMode`** — now `async`; removes the flag before clearing state
- **`login`** — clears the flag automatically when a user signs in, preventing stale guest state
- Updated `App.test.tsx` mock to include the full `AuthContextType` shape (`isGuest`, `continueAsGuest`, `exitGuestMode`, `refreshSession`)

## Behaviour
| Scenario | Before | After |
|---|---|---|
| Cold restart after "Continue as Guest" | Lands on auth screen | Lands on guest home |
| Sign in while in guest mode | Guest flag untouched | Guest flag cleared |
| Exit guest mode | Flag untouched in storage | Flag removed from storage |

## Testing
- All 606 tests pass across 50 suites
- `turbo lint` + `turbo typecheck` clean (pre-existing `socket.io-client` / `react-native-maps` type errors are unrelated)

## Notes
`continueAsGuest` and `exitGuestMode` signatures changed from `() => void` → `() => Promise<void>`. Existing call sites that don't `await` continue to work (fire-and-forget); any that do `await` also correctly.

### DIFF
--- a/apps/mobile/__tests__/App.test.tsx
+++ b/apps/mobile/__tests__/App.test.tsx
@@ -90,10 +90,14 @@ jest.mock('../src/context/AuthContext', () => ({
   AuthProvider: ({ children }: { children: React.ReactNode }) => children,
   useAuth: () => ({
     isAuthenticated: false,
+    isGuest: false,
     isLoading: false,
     user: null,
     login: jest.fn(),
     logout: jest.fn(),
+    continueAsGuest: jest.fn(),
+    exitGuestMode: jest.fn(),
+    refreshSession: jest.fn(),
   }),
 }));
 
--- a/apps/mobile/src/context/AuthContext.tsx
+++ b/apps/mobile/src/context/AuthContext.tsx
@@ -1,5 +1,8 @@
 import React, { createContext, useState, useContext, useEffect, ReactNode } from 'react';
+import AsyncStorage from '@react-native-async-storage/async-storage';
 import { loginWithAzure, logoutFromAzure, loadAuth, saveAuth, AuthResult } from '../auth/AzureAuth';
+
+const GUEST_MODE_KEY = '@SharkPark:isGuest';
 
 export const geofenceLotFilterKey = (email: string) => `@geofence_lot_filter/${email}`;
 
@@ -12,8 +15,8 @@ interface AuthContextType {
   login: () => Promise<void>;
   logout: () => Promise<void>;
   refreshSession: () => Promise<AuthState | null>;
-  continueAsGuest: () => void;
-  exitGuestMode: () => void;
+  continueAsGuest: () => Promise<void>;
+  exitGuestMode: () => Promise<void>;
   isLoading: boolean;
 }
 
@@ -28,10 +31,15 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
   // check for valid token on app launch
   useEffect(() => {
     const initAuth = async () => {
-      const savedUser = await loadAuth();
+      const [savedUser, guestFlag] = await Promise.all([
+        loadAuth(),
+        AsyncStorage.getItem(GUEST_MODE_KEY).catch(() => null),
+      ]);
       if (savedUser) {
         setUser(savedUser);
         setIsAuthenticated(true);
+      } else if (guestFlag === 'true') {
+        setIsGuest(true);
       }
       setIsLoading(false);
     };
@@ -50,6 +58,8 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
 
       await saveAuth(tokens);
       if (__DEV__) console.log('[AuthContext] Tokens saved, setting authenticated');
+      await AsyncStorage.removeItem(GUEST_MODE_KEY);
+      setIsGuest(false);
       setUser(tokens);
       setIsAuthenticated(true);
       if (__DEV__) console.log('[AuthContext] Login complete, isAuthenticated=true');
@@ -101,11 +111,13 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
     return savedUser;
   };
 
-  const continueAsGuest = () => {
+  const continueAsGuest = async () => {
+    await AsyncStorage.setItem(GUEST_MODE_KEY, 'true').catch(() => {});
     setIsGuest(true);
   };
 
-  const exitGuestMode = () => {
+  const exitGuestMode = async () => {
+    await AsyncStorage.removeItem(GUEST_MODE_KEY).catch(() => {});
     setIsGuest(false);
   };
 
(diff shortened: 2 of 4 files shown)


apps/mobile/src/context/AuthContext.tsx
```diff
@@ -28,10 +31,15 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
...
-      const savedUser = await loadAuth();
+      const [savedUser, guestFlag] = await Promise.all([
+        loadAuth(),
+        AsyncStorage.getItem(GUEST_MODE_KEY).catch(() => null),
+      ]);
       if (savedUser) {
         setUser(savedUser);
         setIsAuthenticated(true);
+      } else if (guestFlag === 'true') {
+        setIsGuest(true);
```
chuckmilton, line 50: **Stale-flag edge case**: If a user previously used guest mode, then signed in, the `login` handler clears `GUEST_MODE_KEY` (which is good). However, there's a path where it leaks: user goes guest -> flag set, then somehow signs in via a code path that doesn't go through `login()` (Azure deep-link callback?), then later the auth tokens expire and `loadAuth()` returns null. Now `guestFlag === 'true'` from months ago and the user silently lands on guest home instead of the auth screen.

Mitigation: when `savedUser` exists, opportunistically clear the flag: 
```suggestion
      if (savedUser) {
        setUser(savedUser);
        setIsAuthenticated(true);
        if (guestFlag === 'true') {
          AsyncStorage.removeItem(GUEST_MODE_KEY).catch(() => {});
        }
      } else if (guestFlag === 'true') {
        setIsGuest(true);
      }
```

apps/mobile/__tests__/App.test.tsx
chuckmilton, line 1: The only test change here was updating an existing mock. However, there's no test asserting:
- `initAuth` sets `isGuest=true` when AsyncStorage has the flag and `loadAuth` returns null
- `login` clears the flag
- `continueAsGuest` writes `true` to storage
- `exitGuestMode` removes the key

Given the bug being fixed is exactly "this state didn't survive a restart," a regression test that mocks AsyncStorage and asserts `isGuest` is true after the provider mounts would be cheap and prevent re-regression.

apps/mobile/src/context/AuthContext.tsx
```diff
@@ -50,6 +58,8 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
 
       await saveAuth(tokens);
       if (__DEV__) console.log('[AuthContext] Tokens saved, setting authenticated');
+      await AsyncStorage.removeItem(GUEST_MODE_KEY);
+      setIsGuest(false);
```
chuckmilton, line 72: nit - `login` `removeItem` is unprotected: `continueAsGuest` and `exitGuestMode` defensively wrap `.catch(() => {})`, but here a transient AsyncStorage failure throws after tokens are saved. Azure session is established, but the React state never updates, leaving the user stuck on the loading/auth screen until next restart. We should wrap consistently:
```ts
await AsyncStorage.removeItem(GUEST_MODE_KEY).catch(() => {});
```

apps/mobile/src/context/AuthContext.tsx
```diff
@@ -101,11 +111,13 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
     return savedUser;
   };
 
-  const continueAsGuest = () => {
+  const continueAsGuest = async () => {
```
chuckmilton, line 129: nit - silent `.catch(() => {})` swallows persistence failures: `continueAsGuest`'s catch means a failed write puts `isGuest=true` in memory but not in storage (i.e. the bug this PR fixes silently re-emerges on the next cold start). At minimum log under `__DEV__`:
```ts
.catch((e) => __DEV__ && console.warn('[AuthContext] guest flag persist failed', e));
```

apps/mobile/src/context/AuthContext.tsx
chuckmilton, line 1: nit - `logout` doesn't defensively clear the flag: Logout clears `user`/`isAuthenticated` but not `isGuest`. Today `logout` is only reachable when authenticated (not guest), so this is theoretically dead code. However a one-liner `await AsyncStorage.removeItem(GUEST_MODE_KEY).catch(() => {})` + `setIsGuest(false)` would future-proof against any UI path that ends up calling logout from a guest state.

apps/mobile/src/context/AuthContext.tsx
```diff
@@ -28,10 +31,15 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
...
     const initAuth = async () => {
-      const savedUser = await loadAuth();
+      const [savedUser, guestFlag] = await Promise.all([
+        loadAuth(),
+        AsyncStorage.getItem(GUEST_MODE_KEY).catch(() => null),
+      ]);
       if (savedUser) {
         setUser(savedUser);
         setIsAuthenticated(true);
+      } else if (guestFlag === 'true') {
```
chuckmilton: nit - `'true'` magic string: Use a typed constant or boolean parsing helper:
```ts
const GUEST_FLAG_TRUE = 'true';
// or
const isGuestPersisted = guestFlag === GUEST_FLAG_TRUE;
```

apps/mobile/src/context/AuthContext.tsx
chuckmilton, line 1: nit - Fire-and-forget call sites in `MainTabNavigator.tsx`(unrelated to this PR's scope) - `exitGuestMode()` and `continueAsGuest()` are now async but called without `await`. If the user backgrounds the app immediately after tapping, the AsyncStorage write may not flush. React Native's AsyncStorage usually flushes fast enough that this is fine in practice, but worth either awaiting or adding a `void` prefix for ESLint clarity.

chuckmilton (commented):
Great fix @lawrenceDegoma! Clean and minimal, mirrors the `useOnboarding` pattern exactly which is what we want. The `Promise.all` in `initAuth` is the right move so we don't add a second render pass on cold start.

Couple of small things I'd tighten before merge:

1. Hydration test: the whole point of this PR is "guest flag survives a restart," and there's no test that actually exercises that. Even one spec mocking AsyncStorage + asserting `isGuest === true` after mount would prevent this from silently regressing.

2. Stale-flag edge case in `initAuth`: if `savedUser` exists, clear `GUEST_MODE_KEY` opportunistically. Otherwise an old guest flag from before sign-in could resurface months later when tokens expire.

3. Inconsistency in error handling: `login`'s `removeItem` is unprotected while the other two helpers wrap in `.catch(() => {})`. Worth making them consistent, and probably worth logging under `__DEV__` instead of fully swallowing, since a failed write is exactly the bug we're fixing re-emerging silently.

Rest is super minor (logout could defensively clear the flag, `'true'` could be a constant, fire-and-forget calls in `MainTabNavigator` could `await` or `void`). Once those get revisioned, this will be ready to merge 💯

PR #149, opened 2026-05-02
feat(mobile): wire ReportModal.onSubmit to POST /reports

lawrenceDegoma:
## Summary
Wires `ReportModal.onSubmit` to the live `POST /reports` endpoint (unblocked by PR #121). Previously the submit handler was a no-op (`() => {}`); reports now POST to the backend and surface loading/error state to the user.

### New: `src/services/api/reports.ts`
- `reportsApi.create({ lotId, type, message? })` — POSTs to `/reports`
- `ReportUnauthorizedError` — thrown on 401 (guest / unauthenticated)
- `ReportThrottledError` — thrown on 429 (5/min limit exceeded)
- Both exported from `src/services/api/index.ts`

### `src/services/api/lots.ts`
- Added `id: string` (cuid) to `ParkingLot` interface — the backend already returns it via `...meta` spread in `transformToResponse`; we just weren't declaring it on the mobile type

### `src/components/Modals/ReportModal.tsx`
- `onSubmit` signature: `() => void` → `(report: IncidentReport) => Promise<void>`
- Added `isSubmitting` state — disables submit button and shows `ActivityIndicator` while the request is in flight
- Added `submitError` state — renders an inline error banner (`accessibilityRole="alert"`) on rejection; clears on close
- `handleClose` is a no-op while submitting (prevents closing mid-request)

### `src/screens/ShortTermForecastScreen.tsx`
- `handleReportSubmit` — replaces the `() => {}` stub
  - Guards: guests and unauthenticated users are redirected to the Profile tab (same pattern as the favorites gate)
  - Uses `lot.id` (cuid) as `lotId`, **not** `lot.lot_id` (`'G2'`-style code) — matches the DTO
  - Re-throws `ReportThrottledError` so `ReportModal` surfaces the message; 401 short-circuits to navigation

## Behaviour

| Scenario | Result |
|---|---|
| Authenticated user submits | POST fires, modal closes on success |
| Guest taps Submit | Modal closes, navigates to Profile tab |
| Rate limit hit (5/min) | Inline error banner: *"You're submitting reports too quickly…"* |
| Network / server error | Inline error banner with error message |
| Submit in flight | Button disabled + spinner, backdrop close blocked |

## Testing
- 3 new `ReportModal` tests: correct payload on submit, closes on success, error banner on rejection
- `onSubmit` mock updated to `mockResolvedValue(undefined)` across existing tests
- Test fixtures in `LotAmenities`, `RecommendationModal`, `RealGeofencingIntegration` updated with required `id` field
- **609/609 tests pass** — lint typecheck

### DIFF
--- a/apps/mobile/__tests__/ReportModal.test.tsx
+++ b/apps/mobile/__tests__/ReportModal.test.tsx
@@ -36,7 +36,7 @@ jest.mock('../src/components/CustomTextInput', () => ({
 // ────────────────────── Helpers ──────────────────────
 
 const defaultProps = {
-  lotId: 'G6',
+  lotDisplayName: 'G6',
   isOpen: true,
   onClose: jest.fn(),
   onSubmit: jest.fn().mockResolvedValue(undefined),
@@ -287,7 +287,6 @@ describe('ReportModal -- onSubmit wiring', () => {
     expect(onSubmit).toHaveBeenCalledTimes(1);
     const call = onSubmit.mock.calls[0][0];
     expect(call.type).toBe('blockage');
-    expect(call.lotId).toBe('G6');
     expect(call.timestamp).toBeInstanceOf(Date);
   });
 
@@ -341,5 +340,34 @@ describe('ReportModal -- onSubmit wiring', () => {
     expect(errorView).toBeTruthy();
     const texts = collectTexts(errorView);
     expect(hasText(texts, 'Rate limit exceeded')).toBe(true);
+  });
+
+  it('submit button is disabled while submission is in-flight', async () => {
+    // onSubmit never resolves during this test — simulates a slow network
+    const onSubmit = jest.fn().mockReturnValue(new Promise(() => {}));
+    const tree = render({ ...defaultProps, onSubmit });
+
+    const crashRadio = tree.root.find(
+      node =>
+        node.props.accessibilityRole === 'radio' &&
+        node.props.accessibilityLabel?.includes('Crash'),
+    );
+    await ReactTestRenderer.act(async () => {
+      crashRadio.props.onPress();
+    });
+
+    const submitBtn = tree.root.find(
+      node => node.props.accessibilityLabel === 'Submit report',
+    );
+    // Kick off submit but don't await — leave it in-flight
+    ReactTestRenderer.act(() => {
+      submitBtn.props.onPress();
+    });
+
+    const submitBtnDuring = tree.root.find(
+      node => node.props.accessibilityLabel === 'Submit report',
+    );
+    expect(submitBtnDuring.props.accessibilityState).toMatchObject({ disabled: true });
+    expect(submitBtnDuring.props.disabled).toBe(true);
   });
 });
--- a/apps/mobile/__tests__/ReportModal.test.tsx
+++ b/apps/mobile/__tests__/ReportModal.test.tsx
@@ -39,7 +39,7 @@ const defaultProps = {
   lotId: 'G6',
   isOpen: true,
   onClose: jest.fn(),
-  onSubmit: jest.fn(),
+  onSubmit: jest.fn().mockResolvedValue(undefined),
 };
 
 function render(props = defaultProps) {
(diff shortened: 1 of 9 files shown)


apps/mobile/src/screens/ShortTermForecastScreen.tsx
```diff
@@ -94,6 +95,32 @@ export function ShortTermForecastScreen() {
...
+      navigation.navigate('Profile' as never);
+      return;
+    }
+    if (!lot?.id) throw new Error('Lot data unavailable. Please try again.');
+    try {
+      await reportsApi.create({
+        lotId: lot.id,
+        type: report.type,
+        message: report.message || undefined,
+      });
```
chuckmilton, line 110: `IncidentReport.lotId` is dead data: The modal is passed `lotId={lot.lot_id}` (display code, e.g. `'G2'`) and stuffs that into `IncidentReport.lotId`. But `ShortTermForecastScreen.handleReportSubmit` ignores `report.lotId` and uses `lot.id` (the cuid) directly:
```ts
await reportsApi.create({
  lotId: lot.id,           // cuid — correct
  type: report.type,
  message: report.message || undefined,
});
// report.lotId === 'G2' is silently dropped
```
Two sources of truth, one of them unused. The next person reading the modal contract will assume `report.lotId` is what gets sent.

We should either:
- Drop `lotId` from `IncidentReport` and pass a separate `lotDisplayName: string` prop (only used for the title), **or**
- Pass `lot.id` into the modal as `lotId` and let the screen handler user `report.lotId`.

apps/mobile/src/services/api/reports.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+const reportsApi = {
+  async create(payload: CreateReportPayload): Promise<CreateReportResponse> {
+    try {
+      const response = await apiService.post<CreateReportResponse>(
+        '/reports',
+        payload,
+      );
+      return response.data ?? (response as unknown as CreateReportResponse);
```
chuckmilton, line 53: Response shape should get a comment: This works because `ReportsController.create` returns `{ id, created_at }` raw - not wrapped in `{ success, data }` - while `apiService.post<T>` is typed as `Promise<ApiResponse<T>>`. So `response.data` is `undefined` at runtime and the `??` fallback is the only working path.

That's a pre-existing inconsistency in the codebase, not yours to fix in this PR, but at minimum we should drop a comment explaining why the cast is necessary so the next reader doesn't "clean it up" and break the call. Long-term we should either (a) wrap controller responses consistently or (b) add a typed `apiService.postRaw<T>` helper.

apps/mobile/src/components/Modals/ReportModal.tsx
```diff
@@ -169,20 +183,32 @@ export function ReportModal({ lotId, isOpen, onClose, onSubmit }: ReportModalPro
...
-              accessibilityRole="button"
-              accessibilityLabel="Submit report"
-              accessibilityState={{ disabled: !selectedType || (selectedType === 'other' && !message.trim()) }}
-            >
-              <Text style={styles.submitButtonText}>Submit Report</Text>
-            </TouchableOpacity>
+            {(() => {
+              const isDisabled = !selectedType || (selectedType === 'other' && !message.trim()) || isSubmitting;
+              return (
+                <TouchableOpacity
```
chuckmilton: nit: Hoist `isDisabled` to the top of the component (above `return`) like other derived values. Cleaner JSX, and `isDisabled` becomes reusable if you later want to grey out the message input too.

apps/mobile/src/components/Modals/ReportModal.tsx
chuckmilton, line 1: nit: Modal defines its own literal union `'blockage' | 'crash' | 'other'`, while `services/api/reports.ts` exports `ReportType`. They'll silently drift if either side adds a type. Have the modal import it:
```ts
import type { ReportType } from '../../services/api/reports';
```

apps/mobile/src/components/Modals/ReportModal.tsx
```diff
@@ -169,20 +183,32 @@ export function ReportModal({ lotId, isOpen, onClose, onSubmit }: ReportModalPro
...
+                  disabled={isDisabled}
+                  style={[styles.submitButton, isDisabled && styles.submitButtonDisabled]}
+                  accessibilityRole="button"
+                  accessibilityLabel="Submit report"
+                  accessibilityState={{ disabled: isDisabled }}
+                >
+                  {isSubmitting
+                    ? <ActivityIndicator color={colors.white} />
+                    : <Text style={styles.submitButtonText}>Submit Report</Text>
+                  }
```
chuckmilton: nit: While submitting, the button swaps `<Text>Submit Report</Text>` for just `<ActivityIndicator />`, which is narrower and causes a layout shift. Either keep the text and overlay the spinner, or use `flexDirection: 'row'` to show both.

apps/mobile/__tests__/ReportModal.test.tsx
chuckmilton, line 1: nit: Good coverage on the happy path. Missing the cases most likely to regress:
- Submit button is disabled while `isSubmitting === true`
- `handleClose` is prevented while submitting (backdrop tap is no-op, X button is no-op)
- `submitError` clears when the user retries

apps/mobile/src/screens/ShortTermForecastScreen.tsx
```diff
@@ -94,6 +95,32 @@ export function ShortTermForecastScreen() {
   const [isReliabilityModalOpen, setIsReliabilityModalOpen] = useState(false);
   const [isMapModalOpen, setIsMapModalOpen] = useState(false);
 
+  const handleReportSubmit = async (report: import('../components/Modals/ReportModal').IncidentReport) => {
```
chuckmilton, line 98: nit: No tests for `handleReportSubmit` covering: guest gate, `ReportUnauthorizedError` -> Profile nav, `ReportThrottledError` re-throw, the `!lot?.id` guard. The reports.ts service also has no unit tests.

apps/mobile/src/services/api/reports.ts
```diff
@@ -0,0 +1,60 @@
...
+    } catch (error) {
+      if (error instanceof ApiError) {
+        if (error.status === 401) throw new ReportUnauthorizedError();
```
chuckmilton, line 56: nit: `new ReportThrottledError()` always says `"You're submitting reports too quickly..."`. If the backend later adds a `Retry-After` or a friendlier message in the body, it's swallowed. Consider letting the backend message pass through when present, falling back to the canned string.

chuckmilton (commented):
Great work and clean wiring! The error class hierarchy mirrors the existing `BackgroundLocationRequiredError` pattern nicely, and the loading/disabled UX in the modal is right (love the backdrop-close block during submit). Tests cover the core modal paths.

One real thing to address: `IncidentReport.lotId` is dead data now. The modal builds it from `lot.lot_id` ('G2'-style display code), but `handleReportSubmit` ignores it and sends `lot.id` (cuid) directly. So the field on `IncidentReport` is misleading. Either drop it from the type and pass a separate `lotDisplayName` for the title, or pass `lot.id` into the modal and let it source-of-truth.

Also worth flagging (not blocking): `response.data ?? (response as unknown as CreateReportResponse)` in `reports.ts` is patching around a pre-existing codebase inconsistency, `ReportsController` returns the object raw while `apiService.post<T>` is typed as if it wraps in `{ success, data }`. The cast is actually load-bearing, not dead code, but please drop a comment so the next person doesn't "fix" it and break the call.

Rest is small (IIFE in JSX, `ReportType` drift between modal and api module, missing disabled-while-submitting test, spinner layout shift). Once the `lotId` mismatch is tightened up, we should be ready to merge 🙏

PR #84, opened 2026-04-25
Feature/onboarding deep linking

lawrenceDegoma:
feat(mobile): add onboarding flow, deep linking, and App Store readiness fixes

Onboarding:
- Add useOnboarding hook with AsyncStorage persistence (@SharkPark:onboardingComplete)
- Add OnboardingScreen with 4 slides: welcome (SharkParkV4.webp logo), crowdsource, forecast, permission priming
- Gate onboarding before auth in App.tsx; NavigationContainer wired with linking config
- Add 4 passing tests in useOnboarding.test.ts

Deep linking:
- Add linking.ts with custom scheme (sharkpark://) and universal links (sharkpark.csulb.edu)
- Add iOS URL scheme entry and Associated Domains entitlement (mobile.entitlements)
- Add Android custom scheme and App Links intent filters (autoVerify=true) in AndroidManifest.xml
- Add apple-app-site-association.json stub for server deployment (TEAMID placeholder requires replacement)

App Store readiness fixes:
- Info.plist: fix duplicate UIAppFonts key (was at lines 86 and 123)
- Info.plist: change CFBundleDisplayName from 'mobile' to 'SharkPark'
- Info.plist: remove deprecated NSBluetoothPeripheralUsageDescription (NSBluetoothAlwaysUsageDescription retained)
- Info.plist: replace empty TSLocationManagerLicense with \$(BG_GEOLOCATION_LICENSE) build variable
- AndroidManifest.xml: replace empty locationmanager.license with \${BG_GEOLOCATION_LICENSE} manifest placeholder
- build.gradle: inject BG_GEOLOCATION_LICENSE via System.getenv into manifestPlaceholders
- build.gradle: add production signing config for release builds using env vars (RELEASE_KEYSTORE_PATH, RELEASE_KEYSTORE_PASSWORD, RELEASE_KEY_ALIAS, RELEASE_KEY_PASSWORD); release no longer uses debug keystore
- project.pbxproj: add BG_GEOLOCATION_LICENSE user-defined build setting in Debug and Release configs
- ThemeContext.tsx: wrap console.error calls in __DEV__ guards

MANUAL ACTION REQUIRED before production release:
- Set BG_GEOLOCATION_LICENSE env var in CI/CD (Xcode Cloud / EAS Build / GitHub Actions) with Transistor Software license key
[...]

PR #150, opened 2026-05-02
fix(android): add assetlinks.json for App Links autoVerify

lawrenceDegoma:
## Summary
Adds `assetlinks.json` to unblock Android App Links (post-merge finding from PR #84). `AndroidManifest.xml` already sets `android:autoVerify="true"` on the HTTPS intent filter for `https://sharkpark.app`, but without a valid Digital Asset Links file served at `https://sharkpark.app/.well-known/assetlinks.json`, the OS Domain Verification Service silently fails and Android shows the "Open with…" disambiguation dialog instead of routing directly to the app.

## Changes
- **`apps/mobile/docs/assetlinks.json`** — new Digital Asset Links file
  - Debug SHA-256 fingerprint included (`FA:C6:17:45:DC:09:03:78:...`) — extracted from `android/app/debug.keystore`
  - Release fingerprint placeholder (`RELEASE_SHA256_FINGERPRINT_REPLACE_BEFORE_DEPLOY`) — must be replaced before merge (see blockers below)
- **`apps/mobile/docs/android-app-links.md`** — runbook covering fingerprint extraction, Play App Signing caveat, hosting requirements, and on-device verification with `adb`

## **Blockers** — required before this PR can ship

**This PR is ready for code review but cannot be merged or deployed until both of the following are done:**

**1. Replace the release fingerprint (whoever holds the release keystore)**

Run `keytool` against the release keystore and replace the placeholder in `assetlinks.json`:
```bash
keytool -list -v \
  -keystore <RELEASE_KEYSTORE_PATH> \
  -alias <RELEASE_KEY_ALIAS> \
  -storepass <RELEASE_KEYSTORE_PASSWORD> \
  | grep "SHA256:"
```

If enrolled in Play App Signing: use the fingerprint from Play Console → Release → Setup → App signing → App signing key certificate instead.

Push the updated assetlinks.json as a follow-up commit to this branch before merge.

2. Charles deploys the file to the marketing site

Once merged, the file must be served at:
`https://sharkpark.app/.well-known/assetlinks.json`

Requirements:

Content-Type: application/json
No redirects on the path (Android's verifier follows zero redirects)
The app-side change (android:autoVerify="true") has been live since PR #84 — nothing works end-to-end until this URL is reachable with a valid fingerprint.

## Verification after deploy
```bash
adb shell pm set-app-links --package com.sharkpark.mobile 0 all
adb shell pm verify-app-links --re-verify com.sharkpark.mobile
adb shell pm get-app-links com.sharkpark.mobile  # should show "verified" for sharkpark.app
```


lawrenceDegoma:
I can't access the Google play console to get the app signing key certificate

apps/mobile/docs/assetlinks.json
chuckmilton, line 1: `debug.keystore` is checked into the repo (`.gitignore` line 35: `!debug.keystore`), so anyone who clones can build an APK matching this fingerprint. Once deployed, their APK can claim and intercept all `https://sharkpark.app/*` link taps on a victim's device.

Fix: remove the debug entry from the deployed file. Local devs can verify via sideload/`adb install -r` or a separate `assetlinks.dev.json` that never ships. Update the runbook accordingly: "debug fingerprint = local-only, never deployed."

apps/mobile/docs/assetlinks.json
```diff
@@ -0,0 +1,13 @@
+[
+  {
+    "relation": ["delegate_permission/common.handle_all_urls"],
+    "target": {
+      "namespace": "android_app",
+      "package_name": "com.sharkpark.mobile",
+      "sha256_cert_fingerprints": [
+        "FA:C6:17:45:DC:09:03:78:6F:B9:ED:E6:2A:96:2B:39:9F:73:48:F0:BB:6F:89:9B:83:32:66:75:91:03:3B:9C",
+        "RELEASE_SHA256_FINGERPRINT_REPLACE_BEFORE_DEPLOY"
```
chuckmilton: `"RELEASE_SHA256_FINGERPRINT_REPLACE_BEFORE_DEPLOY"` isn't a valid SHA-256 (must be 32 colon-separated uppercase hex pairs). Android treats malformed entries as a parse failure on the whole statement, deploying as-is silently disables autoVerify, no better than not deploying.

Make it fail loud: use `null`, or add a CI check regex-validating `^[0-9A-F]{2}(:[0-9A-F]{2}){31}$`. Same check would catch the busted `TEAMID` in `apple-app-site-association.json`.

apps/mobile/docs/assetlinks.json
chuckmilton, line 1: With Play App Signing enrolled, only Google's deployment-signing-key fingerprint matters, the upload-key fingerprint from `keytool` is irrelevant because Google re-signs the APK. So the blocker isn't you not being able to access Play Console, it's that I need to pull it from Play Console -> App signing. I'll grab it and push to your branch. Please confirm in the runbook whether we're enrolled in Play App Signing.

apps/mobile/docs/android-app-links.md
chuckmilton, line 1: `AndroidManifest.xml` declares `host="sharkpark.app"` with no `pathPrefix`, so once verified, every path (`/privacy`, `/support`, `/terms`) routes into the app instead of the browser — bad for App Store reviewers needing those pages browsable. Follow-up: scope to `pathPrefix="/lot"`, `/map`, etc.

chuckmilton (commented):
Thanks for unblocking this big man, runbook is solid (Play App Signing call-out especially) and the file structure mirrors `apple-app-site-association.json`

One real blocker before merge: the debug fingerprint can't ship. `debug.keystore` is checked into the repo (gitignore explicitly un-ignores it), so anyone who clones can build an APK matching that fingerprint and hijack `https://sharkpark.app/*` link taps once we deploy. Drop the debug entry and update the runbook to say "debug = local-only, never deployed."

On the Play Console blocker, don't worry about that. With Play App Signing enrolled, the `keytool` upload-key fingerprint is the wrong one anyway; we need Google's deployment-signing-key fingerprint, which only I can pull. I'll grab it and push to your branch. Just confirm in the runbook whether we're on Play App Signing so future-us doesn't get confused.

Smaller stuff: `RELEASE_SHA256_FINGERPRINT_REPLACE_BEFORE_DEPLOY` is silently invalid (Android rejects the whole statement on parse failure), `null` or a tiny CI regex check would fail loud. Same check would catch the busted `TEAMID` in the AASA file. Out-of-scope thought: the intent filter is apex-wide, so every `/privacy` / `/support` tap will route into the app once verified, probably want a path-prefix filter as a follow-up.

Approving once the debug fingerprint comes out ❤️

PR #151, opened 2026-05-02
MLflow R2 Registry

DottL:
# MLflow to R2 model registry upload

Wires --export-s3 to publish the promoted model's artifacts to Cloudflare R2, with a production.json pointer the upcoming Fly prediction job reads to discover the current version.

## Implemented

**Storage layout** in `sharkpark-ml-exports`:

```
s3://sharkpark-ml-exports/models/<model-name>/<version>/
  ├── model.json          # XGBoost median booster
  ├── model_lower.json    # XGBoost 10th-percentile booster (quantile regression)
  ├── model_upper.json    # XGBoost 90th-percentile booster
  └── metadata.joblib     # {category_mappings, feature_columns}

s3://sharkpark-ml-exports/models/<model-name>/production.json
  - {"model_name", "version", "run_id", "promoted_at", "artifact_path", "files"}
```

These four files match exactly what `BaseXGBoostModel.save()` produces `services/ml/src/models/base.py` so the prediction job just calls `BaseXGBoostModel.load(local_dir)` after downloading.

- **`production.json`** is uploaded **last** so the pointer never references a partially-uploaded version. The prediction job reads this file to detect new versions and download only on change. The `files` field lists the artifacts that belong to this version, so the reader iterates the pointer instead of hardcoding its own copy of the filename list.

- **Promotion is idempotent.** Calling `promote --run-id <X>` a second time for the same run reuses the existing registered version instead of creating `v2`, `v3`, … pointing at the same model. 

## Changes
- `services/ml/src/utils/mlflow_utils.py` — adds _upload_to_r2() that downloads the MLflow run's artifacts, uploads them to R2, then writes production.json last. R2 failures log a warning rather than failing the promotion (retry with --upload-only <version\>)
- `services/ml/pyproject.toml` — add `boto3>=1.43.0` (only `botocore` was listed before; insufficient for an S3 client).
- `services/ml/scripts/promote_short_term.py` and `services/ml/scripts/promote_long_term.py` — update `--export-s3` help text.
- `services/ml/uv.lock` — adds `boto3` + transitives (`s3transfer`, bumped `botocore`), and also catches up to the existing `mlflow>=3.11.1` constraint in `pyproject.toml` (lock was pinned to 3.10.1), which pulls in `aiohttp` and its standard dep tree as new transitives. Larger-than-expected diff, but no behaviour change beyond what the constraint already required.
- `services/ml/.env.example` — new file documenting required env vars for ML developers.
- `services/ml/README.md` and `services/ml/Model_Design.md` — updated to reflect the new R2 storage path and remove stale Lambda/S3 references.

## Configuration

| Var | Value | Required? |
| --- | --- | --- |
| `R2_ENDPOINT_URL` | account-scoped R2 S3 endpoint | yes |
| `R2_ACCESS_KEY_ID` | from the R2 API token | yes |
| `R2_SECRET_ACCESS_KEY` | from the R2 API token | yes |
| `R2_BUCKET` | `sharkpark-ml-exports` | optional (default) |

`region_name="auto"` is hardcoded — R2 is globally distributed and ignores region for routing, but boto3 requires the value to compute SigV4 signatures.
## Note
-  **Naming Note**: Provider is **R2**, but identifiers stay **S3** where they're protocol-level (`--export-s3` flag, `boto3.client("s3")`, the S3 wire protocol R2 implements). The new env var is `R2_ENDPOINT_URL` since that *is* R2-specific. Log messages say "R2" so prod logs are unambiguous.
-  **Tracking Server**: MLflow tracking stays **local** for launch. R2 holds only the artifacts the prediction job reads. A shared tracking server on Fly is deferred until a second person needs the dashboard.
- **Old version cleanup / retention.** Every promoted version stays in R2. Storage and egress costs are negligible at the scale we're operating; revisit if that changes (Total ~4 MB per version)


services/ml/src/utils/mlflow_utils.py
```diff
@@ -44,20 +57,42 @@ def promote_model(
...
-        return None, False, None
+        registered_versions = client.search_model_versions(f"name='{model_name}'")
+    except mlflow.exceptions.MlflowException:
+        registered_versions = []
 
-    version = result.version
+    existing = next(
+        (mv for mv in registered_versions if run_id in (mv.source or "")),
+        None,
+    )
```
chuckmilton, line 88: Substring match on `mv.source`. MLflow run IDs are 32-hex so collision is improbable in practice, but the proper field is `mv.run_id`, fall back to parsing only when it's None (you already have `_resolve_run_id` for exactly this):

```suggestion
    existing = next(
        (mv for mv in registered_versions if _resolve_run_id(mv) == run_id),
        None,
    )
```

Same logic, no substring ambiguity, reuses the helper.

services/ml/src/utils/mlflow_utils.py
chuckmilton, line 1: Re-promoting silently moves `@production` backwards: The idempotent reuse runs before `set_registered_model_alias`, so calling `--promote --run-id <old_run>` when `<old_run>` is already registered as `v3` and `@production` currently points at `v5` will silently move `@production` back to `v3`. That might be an intentional rollback, or it might be a footgun ("I wanted to re-export v3's artifacts, not change production").

Two mitigations to choose from:
- Document the behavior explicitly in the `promote_model` docstring ("re-promoting reuses the existing version AND re-points the alias")
- Log a warning when `existing.version != current_production_version` so the operator notices
- Or split: `--upload-only` for re-export without alias change (which is what the new flag already kind of does), and require an explicit `--force-realias` to move `@production` backwards

services/ml/tests/utils/test_mlflow_utils.py
```diff
@@ -0,0 +1,156 @@
...
+    def test_missing_r2_env_vars_does_not_fail_promotion(
+        self, trained_run_id, monkeypatch
+    ):
+        """Missing R2 credentials log a warning but mlflow promotion still succeeds."""
+        # Ensure no R2 env vars are set
+        for var in ("R2_ENDPOINT_URL", "R2_ACCESS_KEY_ID", "R2_SECRET_ACCESS_KEY"):
+            monkeypatch.delenv(var, raising=False)
+
+        version = promote(trained_run_id, export_s3=True)
+        assert version is not None
```
chuckmilton, line 119: nit: Test only asserts `version is not None`. It would still pass if someone deleted the env-var check entirely and the upload silently no-op'd against a broken client. Add a `caplog` assertion that the "R2 upload failed" warning was logged with the expected env-var-missing message:
```suggestion
    def test_missing_r2_env_vars_does_not_fail_promotion(
        self, trained_run_id, monkeypatch, caplog
    ):
        """Missing R2 credentials log a warning but mlflow promotion still succeeds."""
        # Ensure no R2 env vars are set
        for var in ("R2_ENDPOINT_URL", "R2_ACCESS_KEY_ID", "R2_SECRET_ACCESS_KEY"):
            monkeypatch.delenv(var, raising=False)
        with caplog.at_level(logging.WARNING):
            version = promote(trained_run_id, export_s3=True)
        assert version is not None
        assert any("Missing required R2 env vars" in r.message for r in caplog.records)
```

services/ml/src/utils/mlflow_utils.py
```diff
@@ -74,14 +109,144 @@ def promote_model(
...
+
+    Required env vars: R2_ENDPOINT_URL, R2_ACCESS_KEY_ID, R2_SECRET_ACCESS_KEY.
+    Optional: R2_BUCKET (defaults to ``sharkpark-ml-exports``).
+    """
+    import boto3
+
+    endpoint_url = os.environ.get("R2_ENDPOINT_URL")
+    access_key = os.environ.get("R2_ACCESS_KEY_ID")
+    secret_key = os.environ.get("R2_SECRET_ACCESS_KEY")
+    bucket = os.environ.get("R2_BUCKET", _DEFAULT_R2_BUCKET)
```
chuckmilton: nit: If someone sets `R2_BUCKET=` (empty) in `.env`, you get an empty bucket name and S3 calls fail with a confusing error. Use `or`:
```suggestion
    bucket = os.environ.get("R2_BUCKET") or _DEFAULT_R2_BUCKET
```

services/ml/src/utils/mlflow_utils.py
```diff
@@ -74,14 +109,144 @@ def promote_model(
...
+        for filename in _MODEL_ARTIFACT_FILES:
+            src = local_path / filename
+
+            if not src.exists():
+                raise FileNotFoundError(
+                    f"Required artifact {filename} missing from run {run_id}"
+                )
+
+            key = f"{artifact_prefix}/{filename}"
+            client.upload_file(str(src), bucket, key)
```
chuckmilton: nit: `client.upload_file(...)` sends files as `binary/octet-stream` by default. The prediction job won't care (downloads by filename), but anyone debugging via `curl` against R2 directly gets non-rendered JSON. Cheap improvement:
```suggestion
            content_type = "application/json" if filename.endswith(".json") else "application/octet-stream"
            client.upload_file(str(src), bucket, key, ExtraArgs={"ContentType": content_type})
```

services/ml/src/utils/mlflow_utils.py
```diff
@@ -15,8 +18,18 @@
     "load_run_data",
     "get_production_run_id",
     "promote_model",
+    "upload_model_to_r2",
 ]
 
+# Matches BageXGBoostModel.save() in services/ml/src/models/base.py
```
chuckmilton: nit: small typo in comment
```suggestion
# Matches BaseXGBoostModel.save() in services/ml/src/models/base.py
```

services/ml/src/utils/mlflow_utils.py
```diff
@@ -74,14 +109,144 @@ def promote_model(
...
+            _upload_to_r2(model_name=model_name, version=version, run_id=run_id)
+        except Exception as e:
+            logger.warning(
+                "R2 upload failed for %s v%s — model is registered locally but "
+                "not yet published to R2. Retry with --upload-only %s. (%s)",
+                model_name,
+                version,
+                version,
+                e,
+            )
```
chuckmilton, line 149: nit (bare `except Exception` on R2 catch): Intentional per the PR description, and `Exception` doesn't catch `KeyboardInterrupt`/`SystemExit` so the blast radius is bounded, but tightening to the actual surface area (`RuntimeError` for env-var failures, `FileNotFoundError` for missing artifacts, `botocore.exceptions.BotoCoreError`/`ClientError` for uploaded failures) would surface programming bugs immediately instead of swallowing them.

This is optional tho, current code is defensible.

services/ml/src/utils/mlflow_utils.py
chuckmilton, line 1: nit (concurrent-promotion race (future concern, not blocking):
`put_object` for `production.json` isn't conditional. Two concurrent `promote --export-s3` invocations could result in the pointer referencing version A while a partial v B upload was in progress. With manual-script promotion this is essentially impossible, flagging only as a "when this moves to a scheduled job" follow-up. R2 supports `If-Match`/`If-None-Match` for this.

chuckmilton (commented):
Very solid implementation @DottL! The careful bits (pointer-last write, idempotent re-promotion, separate `--upload-only` path that fails loud while inline upload swallows) are exactly right and the upload-ordering test is the kind of thing that catches silent regressions later. `.env.example` and the naming-note in the description are super helpful too.

Two real things to tighten before merge:

1. Idempotency check uses `run_id in (mv.source or "")` - substring match. Run IDs are 32-hex so collision is improbable, but the right field is `mv.run_id` (with `_resolve_run_id` as the existing fallback for when it's None). Quick swap, no ambiguity.
2. Re-promoting moves `@production` backwards silently. If `v3` is registered for run X and `@production` currently points at `v5`, calling `promote --run-id <X>` reuses `v3` AND re-aliases. Could be intentional rollback or a footgun. Either document explicitly in the docstring, or log a warning when the reused version isn't already the production one.

Smaller stuff: `test_missing_r2_env_vars_does_not_fail_promotion` only asserts `version is not None` - would pass even if the env-var check was removed entirely. Add a `caplog` assertion. `R2_BUCKET=""` empty-string would override the default - use or `_DEFAULT_R2_BUCKET`. `BageXGBoostModel`  typo on the comment above `_MODEL_ARTIFACT_FILES`. And one future thought (not this PR): when this moves to a scheduled job, the pointer `put_object` should be conditional (`If-Match`) - concurrent promotes could otherwise point at v A with v B's files mid-upload. Manual-script-only today, no risk.

Once the main issues are taken care of, we should be all good to merge 💯

chuckmilton (approved):

PR #160, opened 2026-05-03
feat(mobile): force-update screen on launch via GET /min-version

lawrenceDegoma:
## Summary
Adds a blocking force-update screen that fires on every app launch. If the device's installed version is below the server-enforced minimum, the user sees a full-screen prompt and cannot proceed until they update.

### `src/services/api/version.ts` *(new)*
- `fetchMinVersion()` — calls `GET /min-version` and returns `{ minSupportedVersion: string }`
- Fails with a standard `ApiError` on non-2xx (handled by the existing `apiService` retry/error layer)

### `App.tsx`
- `semverLt(a, b)` — lightweight major.minor.patch comparator (no external dependency)
- `useForceUpdate()` — hook that calls `fetchMinVersion` on mount, compares against `DeviceInfo.getVersion()`, and sets `updateRequired`
- **Fail-open**: network errors are caught and logged in `__DEV__`; the user is never incorrectly blocked on a flaky connection
- Gates `AppContent` render: shows `null` until the version check resolves, then `<ForceUpdateScreen />` if outdated

### `src/screens/ForceUpdateScreen.tsx` *(new)*
- Full-screen blocking UI — no back/dismiss affordance
- SharkPark logo (`200×120`, same dimensions as `LoginScreen`)
- Platform-aware CTA: **"Update on App Store"** (iOS) / **"Update on Play Store"** (Android)
- Store URLs defined as constants at the top — swap in the real App Store ID once published

### `src/screens/index.ts`
- Exports `ForceUpdateScreen`

### `__tests__/ForceUpdateScreen.test.tsx` *(new)*
4 specs covering the gate logic in `App.tsx`:
| # | Scenario | Expected |
|---|----------|----------|
| 1 | `currentVersion < minSupportedVersion` | Force-update screen shown |
| 2 | `currentVersion === minSupportedVersion` | Normal app renders |
| 3 | `currentVersion > minSupportedVersion` | Normal app renders |
| 4 | `fetchMinVersion` throws (network error) | Fail-open, normal app renders |

## Testing
```bash
pnpm --filter mobile test --no-coverage
# 614 tests, 0 failures

### CODE REVIEW

**Comment by chuckmilton** [comment]
## Review
Thorough pass — overall solid: clean separation (`version.ts` API, hook in App.tsx, screen file), fail-open is the right default, and the 4 gate-logic tests are clear. I pushed `6de1dff` directly to this branch with the fixes below so we don't bounce another revision back to you.

### Issues found + fixes pushed

1. **🔴 Wrong Android package ID (would 404 on Play Store).**
   `STORE_URLS.android` used `com.sharkpark.mobile`, but the real `applicationId` in `android/app/build.gradle` is `app.sharkpark.mobile` (`namespace` is `com.mobile`, which is unrelated to the Play Store URL). Fixed.

2. **🟡 `<ForceUpdateScreen />` rendered without `SafeAreaProvider`.**
   The screen uses `SafeAreaView` from `react-native-safe-area-context`, which falls back to a no-inset frame (and warns) when there's no `SafeAreaProvider` ancestor. Every other branch in `AppContent` wraps in `<SafeAreaProvider>` + `<StatusBar>` — I made the `updateRequired` branch do the same.

3. **🟡 `semverLt` mis-handled short versions.**
   `parseInt('') || 0` works for `0`, but `[aMaj, aMin, aPatch] = parse('1.0')` left `aPatch = undefined`, and `undefined < 0` is `false`. Refactored to default each missing slot to `0`. Not a bug today (we always emit 3-part semvers), but cheap insurance.

4. **🟡 Codecov patch coverage 50% — added direct render tests.**
   Existing test mocks out `ForceUpdateScreen` entirely, so `handleUpdate` + render path were uncovered. Added [`__tests__/ForceUpdateScreen.render.test.tsx`](apps/mobile/__tests__/ForceUpdateScreen.render.test.tsx) (3 specs):
   - iOS: CTA copy + `Linking.openURL` is called with App Store URL on press
   - Android: CTA copy + asserts the **corrected** `app.sharkpark.mobile` Play Store URL
   - Title + body copy renders

   The Android assertion would have caught issue #1.

### Things noted but not changed

- **iOS App Store URL placeholder `id0000000000`** — fine to leave with the `TODO` comment; iOS publishing isn't done yet. Tracked in TODO.md under "Pre-launch credentials & signing." We'll swap it once App Store Connect assigns the ID.
- **No re-check on `AppState` foreground** — currently we only check on mount. If a user has the app backgrounded for days during a force-update rollout, they won't see the screen until next cold start. Acceptable for v1; can add later if needed.
- **Depends on backend [#175](https://github.com/SharkPark-App/SharkPark/pull/175)** — until that ships, every launch hits 404 → fail-open path → users proceed normally. Confirmed `apiService` will throw `ApiError(404, ...)`, the hook catches it, no false blocks.

### Test results
```
Test Suites: 2 passed, 2 total
Tests:       7 passed, 7 total
```
TypeScript clean (`tsc --noEmit`).

Once CI is green I'll squash-merge. Thanks @lawrenceDegoma 🦈

PR #163, opened 2026-05-03
feat(backend): analytics endpoints

zachfpadilla:
# Lot Analytics Endpoints
## Summary
Adds two new `/lots/` endpoints - one for returning the hourly occupancy averages for the past # of days & one for returning per-lot utilizations rates.

### Trends Endpoint: `GET /api/v1/lots/:id/trends?range=<num>d`
- returns hourly occupancy averages for the past `<num>` days
  - Utilizes raw `occupancy_snapshots` data
- Clamps inputs to `[1, 90]`, and silently defaults on unrecognized input (avoids breaking system due to bad queries)

### Utilization Endpoint: `GET /api/v1/lots/utilization?range=<num>d`
- returns per-lot utilization rates for the past `<num>` days
- joins `lot.findMany` + `occupancySnapshot.groupBy` - lots with no utilization return `avg_utilization: null` rather than `0`
  - differentiates between empty lots vs. simply no data

## Notes
- The response doesn't include the actual date window; could either be handled by returning `since` and `until` ISO strings, or will just have to be calculated by the requesting client themselves
- The trends endpoint uses raw `occupancy_snapshots` data - if this endpoint were to be used for measuring actual lot occupancies, it would likely be better to average `estimated_occupancy / capacity`
- Returned timestamps are in UTC


chuckmilton:
## Review — analytics endpoints

Merged `origin/main` into the branch and resolved conflicts in `apps/backend/src/lots/lots.{service,controller.spec}.ts` (combined the new `Lot`/`Prisma`/`LotType` imports + kept all three controller `describe` blocks: `getLotTrends`, `getUtilization`, `getNearbyEvents`). Pushed as 9cbacf5.

### What I checked

- **SQL safety** — `getTrends` raw query uses Prisma tagged-template parameterisation for `lot.id` and `since`. ✅
- **Index coverage** — `idx_snapshots_lot_time(lot_id, timestamp)` covers both queries. ✅
- **Auth tier** — `/lots/utilization` and `/lots/:id/trends` are public, consistent with the locked-Public decision for `/lots/:id/history` (aggregates, no live signal). ✅ See `docs/api-access-tiers.md`.
- **Cache headers** — `max-age=300, s-maxage=600` matches `/history`. ✅
- **Range parsing** — `parseRangeDays` clamps `[1, max]`, defaults silently on invalid input. Edge cases (`'0d'`, `'200d'`, `'1week'`) covered in unit tests. ✅
- **Null handling** — `avg_utilization: null` for lots with zero snapshots in window; sort puts them last via `?? -1`. ✅
- **Response shape** — `{success, range_days, count, data}` is consistent with sibling endpoints. ✅
- **Tests** — 70/70 pass after merge resolution. `tsc` clean. `eslint` clean.

### Non-blocking suggestions (future PR)

- Consider adding `since`/`until` ISO timestamps to the response so clients can label/cache without recomputing the window.
- `date_trunc('hour', timestamp)` truncates in the DB session timezone; document explicitly that buckets are UTC (matches how Postgres on Neon is configured).

LGTM once CI is green.

chuckmilton:
Implemented both suggestions in d17e71b:

- `/lots/utilization` and `/lots/:id/trends` now return `since` + `until` ISO timestamps. Window is computed once in the controller from a single `new Date()` so the response is internally consistent.
- Added a doc comment on `getTrends` clarifying that `date_trunc('hour', timestamp)` buckets are UTC (Postgres on Neon session TZ).

Spec asserts `until - since === rangeDays * 86400000`. 70/70 unit + lint + tsc clean.

PR #173, opened 2026-05-03
feat(weather): migrate to NWS api.weather.gov, add long-term forecast cron, prune weather observations

chuckmilton:
## Summary

Replaces OpenWeather (OWM) with the keyless **US National Weather Service** API across the observation cron, and introduces a new 6-hourly **forecast** cron that populates a new `weather_forecasts` table with hourly NWS predictions for ML's long-term horizon.

## Why

OWM's free `/data/2.5/weather` endpoint never returned a precipitation probability, so `weather.precipitation_probability` had been silently `0` in production — `WeatherService`'s demand-impact branch (`> 0.6`) and ML's HEAVY_RAIN check (`> 0.7`) could never fire. NWS hourly forecast carries a real `probabilityOfPrecipitation`, so swapping vendors fixes that bug *and* unlocks a forecast pipeline at the same time. NWS is also free, key-less, and US-government-operated — no quota or billing risk.

## Backend changes

- **`NwsClient`** (`apps/backend/src/weather/nws.client.ts`) — two-step `/points` → `/gridpoints/.../forecast/hourly`. 24h in-process cache for the point-resolution call (gridpoints rarely change). Helpers exported for unit testing:
  - `parseWindSpeedMph` — handles `"5 mph"` and `"5 to 10 mph"`
  - `probabilityToRate` — null-safe percent → 0..1
  - `deriveIsRaining` — rain keyword **AND** pop ≥ 0.5 (avoids "Slight Chance Rain Showers" false positives)
  - `computeFeelsLikeF` — Rothfusz heat-index (T ≥ 80°F & RH > 40%) + Osczevski wind-chill (T ≤ 50°F & wind > 3 mph), otherwise dry-bulb
- **`WeatherFetchService`** rewritten on top of `NwsClient`. Same `Weather` schema (no migration), now persists a real `precipitation_probability` and a derived `feels_like_f`.
- **New `WeatherForecastFetchService` + `fetch-weather-forecast` cron** (`0 */6 * * *`) writes ~156 hourly rows per run, upserted on `(school_id, target_time)`. Past-time rows are pruned each run, so `weather_forecasts` is self-bounded.
- **New `weather_forecasts` Prisma model + migration** (`20260503111924_add_weather_forecasts`).
- **`pruneOldData` extended** to also delete `weather` observations older than 30 days. ML reads only the latest weather row, so historical observations become dead weight after the 3h staleness gate. `OccupancySnapshot` rows remain permanent (training source).
- **Sentry Cron monitor** for `fetch-weather-forecast` registered in `_cron-monitors.ts` (auto-registers on first check-in; lockstep test enforces parity with `crontab`).

## Config / ops

- `OPENWEATHER_API_KEY` removed from `configuration.ts`, `.env.example`, runbook secrets table, and `api-access-tiers.md`.
- `WEATHER_USER_AGENT` added (NWS requires a descriptive UA so they can contact heavy users). Default in code: `SharkPark/1.0 (ops@sharkpark.app)` — already routed via Cloudflare Email Routing to my inbox, so override is optional.
- `WEATHER_LAT` / `WEATHER_LON` added with CSULB defaults.

**Post-merge ops follow-up:**
```bash
flyctl secrets unset OPENWEATHER_API_KEY -a sharkpark-api
# WEATHER_USER_AGENT default is correct; only set if changing contact mailbox
```

## ML compatibility

No code change required. `services/ml/src/data/loader.py::fetch_latest_weather` still reads the same `weather` columns (`temperature_f`, `humidity_pct`, `precipitation_probability`, `is_raining`, `recorded_at`) via `LIMIT 1`, and the `WeatherSnapshot` field contract is preserved. `weather_adjustment.py` keyword comment refreshed (was OWM-specific).

⚠️ **Behavior change (intended):** the HEAVY_RAIN / `> 60%` pop branches in `weather_adjustment.py` will now actually fire — they were dead code under OWM. Kill-switch `WEATHER_ADJUSTMENT_ENABLED=false` available if any regression appears in production.

## Tests

- **129/129 passing** across `weather`, `config`, `scripts`, `occupancy-events`.
- **New:** `nws.client.spec.ts`, `weather-forecast-fetch.service.spec.ts`.
- **Updated:** `weather-fetch.service.spec.ts` (mocks `NwsClient`), `occupancy-events.service.spec.ts` (`pruneOldData` covers `weather` table), `configuration.spec.ts` (`weather.userAgent`), `_cron-monitors.spec.ts` (lockstep with crontab).
- ML: 20/20 passing on the unchanged Python suite.

## Follow-ups (separate PRs)

- Wire `weather_forecasts` into the long-term predict pipeline on the ML side.


services/ml/src/postprocess/weather_adjustment.py
```diff
@@ -70,8 +70,10 @@ class WeatherSnapshot:
 # current values are a hand-picked heuristic, not derived.
 _COMMUTE_HOURS = (7, 8, 9, 16, 17, 18)
 
-# Keywords matched against OWM's text.
-# "freezing rain" (OWM id 511, Rain group) is bucketed as snow.
+# Keywords matched against the lower-cased NWS `shortForecast` text written
+# by the backend's WeatherFetchService (e.g. "heavy rain", "scattered
+# thunderstorms", "freezing rain likely"). Freezing rain is bucketed as snow
+# because its impact on driver behavior matches snow more than rain.
 _SEVERE_KEYWORDS = ("thunderstorm", "tornado", "squall")
```
DottL, line 77: (non-blocking) NWS frequently emits low-probability thunderstorm strings like "Slight Chance Showers And Thunderstorms" and "Chance Showers And Thunderstorms". 
- The ML classifier in services/ml/src/postprocess/weather_adjustment.py matches the bare thunderstorm keyword and immediately classifies as SEVERE (50% median reduction), which would over-correct on a 20% chance forecast 5 days out.
- I'll roll the fix into the long-term weather-aware PR. Flagging for documentation sake.

TODO.md
DottL, line 228: Did you want to move both cron tasks up to your section since you're already updating docs🙏 🙏 
- Also, no backfill is needed since the prediction scripts already emit all future hours on every run.

apps/backend/src/weather/nws.client.ts
```diff
@@ -0,0 +1,253 @@
...
+    return {
+      updateTime: data.properties.updateTime,
+      periods: data.properties.periods,
+    };
+  }
+
+  /**
+   * Clear the point cache. Test seam — production code should never need this.
+   */
+  resetCache(): void {
```
zachfpadilla: nit: As this is only meant to be used for testing, maybe you could prefix it with `_` to signify that it's internal? (e.g. `_resetCache()`)

zachfpadilla (approved):
Everything looks good to me - commented on one major issue though, be sure to take a look at it.

apps/backend/src/scripts/prune-old-data.ts
```diff
@@ -2,12 +2,13 @@ import { runCronJob } from './_bootstrap';
...
 
 /**
- * Raw-event data retention cron.
+ * Raw-data retention cron.
  *
- * Deletes rows from `occupancy_events` older than 30 days, honoring the
- * user-facing privacy promise in README.md. `occupancy_snapshots` and
- * all aggregated tables are kept permanently (see infrastructure/README.md
- * "Data Retention") — this script never touches them.
+ * Deletes rows older than 30 days from `occupancy_events` (privacy
```
DottL: prune call weather is removed, update doc

DottL (approved):
🌻

apps/backend/src/weather/nws.client.ts
chuckmilton: Good catch @zachfpadilla, can't believe I missed that (almost embarrassing tbh). I might get this fixed next week due to the concern that I might have to restructure the entire architecture of our app. Sorry if this caused an inconvenience on your end whatsoever 🙏

apps/backend/src/scripts/prune-old-data.ts
chuckmilton: [My reaction](https://tenor.com/view/guy-punches-sunflowers-pvz-plants-vs-zombies-gif-6137659021579759562)

PR #175, opened 2026-05-03
feat(backend): min version endpoint

zachfpadilla:
# Minimum Version Endpoint
## Summary
Adds a `GET /api/v1/min-version` endpoint. This will be invoked upon app launch to check if the user needs to update their app before being permitted to access it.

### `GET /api/v1/min-version`
- Hardcoded to return:
```{ ios: { min: "1.0.0", current: "1.0.0" }, android: { min: "1.0.0", current: "1.0.0" } }```
  - Can eventually be moved to `.env`, or into a `MobileVersion` Prisma model
- Invoked by @lawrenceDegoma's force-update screen upon every app launch
  - See: #160

### CODE REVIEW

**Review by chuckmilton — CHANGES_REQUESTED** [review]
Thanks Zach — endpoint is small and clean, but there's a **breaking contract mismatch with the mobile force-update screen (PR #160) that already merged into `main`**. As written, this PR will silently disable force-update for every user once it ships, because the mobile gate fail-opens on a parse miss.

### Blocker — response shape doesn't match the mobile contract

Mobile (`apps/mobile/src/services/api/version.ts`, already on `main`) does:

```ts
export interface MinVersionResponse {
  minSupportedVersion: string;
}

export async function fetchMinVersion(): Promise<MinVersionResponse> {
  const response = await apiService.get<MinVersionResponse>('/min-version');
  return response.data; // <-- the unwrapped `data` payload
}
```

`apiService.get<T>` already strips the outer `{ success, data }` envelope, so it expects the inner `data` to be `{ minSupportedVersion: string }`. This PR returns:

```jsonc
{ "success": true, "data": { "ios": { "min": "1.0.0", "current": "1.0.0" }, "android": { "min": "1.0.0", "current": "1.0.0" } } }
```

Result: `response.data.minSupportedVersion` is `undefined`, `semverLt(deviceVersion, undefined)` throws, App.tsx fail-open swallows it, and no user is ever force-updated. Worst-case silent break.

**Fix options (pick one):**

1. **Match the existing mobile contract (smallest change, recommended for this PR):** return
   ```ts
   { success: true, data: { minSupportedVersion: '1.0.0' } }
   ```
   Drop the per-platform / `current` fields for now. We can extend later when there's a real divergence between iOS/Android minimums; YAGNI for v1 since both store binaries ship together.

2. **Keep your richer per-platform shape, but update mobile in the same PR** so the contract stays consistent and `MinVersionResponse` reads platform-specific (e.g. `Platform.OS === 'ios' ? data.ios.min : data.android.min`). Heavier, requires touching the mobile app + tests.

I'd go with option 1 unless you actively need separate iOS/Android floors before launch.

### Nits / non-blocking

- `current` is unused by the only consumer (mobile force-update gate). Either drop it or document what it's for in a code comment so the next person doesn't think it's dead.
- The hardcoded version is fine for landing the endpoint, but please open a follow-up issue (or note in PR body) to back this with `process.env.MIN_SUPPORTED_VERSION_IOS` / `_ANDROID` Fly secrets before we cut a v1.1 — otherwise bumping the floor requires a backend deploy. Prisma `MobileVersion` model is overkill until we want admin UI to bump it; env vars are sufficient short-term.
- Test currently asserts `expect.any(String)` — once the shape is fixed, please assert the literal `'1.0.0'` so a future accidental shape change is caught. Right now the test would still pass if `min` got renamed to `minimum`.
- `MinVersionModule` doesn't strictly need to exist — controller could be registered directly on `AppModule`. Keeping the module is fine (matches house style) just flagging.

Once the response shape is aligned, this is good to go.

chuckmilton:
Pushed `a55f358` directly to the branch with the contract fix and a couple of related improvements — flagging here so you can scan and re-request review when ready:

**Changes**
- `min-version.controller.ts` — response now `{ success: true, data: { minSupportedVersion: string } }` to match the mobile contract that PR #160 already shipped on `main`.
- New `min-version.service.ts` — reads `MIN_SUPPORTED_APP_VERSION` env var (strict `MAJOR.MINOR.PATCH`), defaults to `'1.0.0'` with a boot warn when unset, **throws** on a malformed override (per the no-graceful-skip principle: a typo in the release-day bump should fail the deploy, not silently serve the old default and leave users un-force-updated).
- `min-version.module.ts` — wires the service.
- `min-version.controller.spec.ts` — replaced the `expect.any(String)` assertion with three specs covering default, env override, and malformed-input throw.
- `.env.example` — documents `MIN_SUPPORTED_APP_VERSION` under Optional.

**What's next on the path to actually using this**
- Set `MIN_SUPPORTED_APP_VERSION` as a Fly secret on `sharkpark-api` before any release where we actually want to enforce a floor. Until then the `'1.0.0'` default means no one is force-updated (mobile is on `1.0.0`).
- Per-platform floors (separate iOS/Android minimums) intentionally deferred — mobile doesn't send a `Platform` header today, and we ship both store binaries from the same RN build with the same version, so YAGNI for v1. If we ever need it, we'd add `MIN_SUPPORTED_APP_VERSION_IOS` / `_ANDROID` env vars and have mobile send `x-platform: ios|android` so the controller can pick the right floor server-side.

Local: typecheck + lint + min-version tests all green.

chuckmilton:
Pushed `3e520d3` — per-platform floors are now built in to this PR.

**Backend**
- `MinVersionService` resolves `per-platform → global → '1.0.0'` and reads two new optional env vars: `MIN_SUPPORTED_APP_VERSION_IOS` and `MIN_SUPPORTED_APP_VERSION_ANDROID`. Each is independently semver-validated and fails loud on a malformed value (typos at deploy time crash the boot rather than silently serving the previous floor).
- `MinVersionController` reads the `x-platform` request header and passes it to the service. **Response shape is unchanged** (`{ minSupportedVersion }`) — the platform-ness is server-side only, so we can evolve resolution logic later without a coordinated mobile release.
- Spec is now 10 cases: global default/override/malformed, per-platform iOS+Android, case-insensitive header matching, partial-override fallback, missing/unknown header fallback, and per-platform malformed → fail-loud.
- `.env.example` documents both new vars.

**Mobile**
- `API_CONFIG.DEFAULT_HEADERS` now includes `x-platform: Platform.OS` so it rides on every request automatically — no per-call wiring needed. At v1 launch this is 100% of installs (both stores ship from the same RN bundle and the header lands in the very first build users install). Old clients without the header get the global floor, which is always the safe lower bound.
- `api.config.test.ts` updated.

**Operational notes**
- For launch: leave both per-platform vars unset and either leave `MIN_SUPPORTED_APP_VERSION` unset (boot warn, defaults to `'1.0.0'`) or set it to `'1.0.0'` explicitly on Fly. Mobile is shipping `1.0.0`, so no one gets force-updated on day 1 — correct for a first release.
- First time a store-approval skew bites us: `fly secrets set MIN_SUPPORTED_APP_VERSION_IOS=<floor> -a sharkpark-api` (or `_ANDROID`) without a code deploy.

All green locally: backend min-version 10/10, mobile api.config 9/9, mobile api.service 15/15, both apps typecheck + lint clean.

PR #178, opened 2026-05-03
feat(backend): GDPR/CCPA data export endpoint

zachfpadilla:
# GDPR/CCPA Data Export Endpoint

## Summary

Adds a data-export endpoint required for GDPR Article 20 (right to data portability) and CCPA compliance, and for accurate App Store privacy answers. Pairs with the existing `DELETE /users/me` to complete the privacy section.

The endpoint returns a single JSON document containing everything the platform holds on the authenticated user: profile fields, favorited lots, push tokens, user-submitted reports, and notification delivery logs. Every call is audited via the existing `audit_events` table using the same hashed-actor pattern as account deletion — no reversible PII is written.

### `GET /api/v1/users/me/data`

- Route declared before `GET :userId` in the controller so NestJS does not shadow it with the dynamic segment.
- Auth is derived from the JWT claim (`req.user.email`), identical to `DELETE /me` — no `:userId` param means no IDOR surface.
- Returns HTTP 200 with `{ success: true, data: UserDataExport }` or 403 if the token carries no email.

### Response shape

```jsonc
{
  "success": true,
  "data": {
    "exported_at": "<ISO timestamp>",
    "profile": {
      "email", "first_name", "last_name", "user_type",
      "phone", "notification_preferences", "created_at", "last_login"
      // internal `id` is intentionally omitted
    },
    "favorites":         [{ "lot_id", "added_at" }],
    "push_tokens":       [{ "token", "platform", "registered_at" }],
    "reports":           [{ "lot_id", "type", "message", "submitted_at" }],
    "notification_logs": [{ "type", "lot_id", "sent_at" }]
  }
}
```

All lot references are the human-readable `lot_id` (e.g. `"G1"`), not internal UUIDs.

### Audit trail

Every successful export writes a `USER_DATA_EXPORTED` row to `audit_events` (enum value already existed in the schema). The `actor_hash` is `SHA-256(salt:email)` — same salt and format as `USER_DELETED` rows — so the audit log proves an export happened without storing reversible PII.

### Privacy/Data Inventory Document (`docs/privacy-data-inventory.md`)
To be used for filling out the App Store privacy labels questionnaire and Play Store data safety form.

**Potential verified issues (can be resolved in separate PR)**
- **`notification_logs` retention** — no prune cron exists yet. Rows are small but will grow indefinitely.
  - **Potentional solution**: 90-day window to match Sentry's crash log retention.
- **`contributor_pings` retention** — no prune cron. Once a device stops contributing its row lives forever.
  - **Potentional solution**: Periodic cleanup, but data is low-risk (pseudonymous).
- **Phone number** — column exists but is not collected in any current flow.
  - **Potentional solution**: Not currently a problem. If it is ever collected, however, then it must be added to the App Store label under Contact Info → Phone Number.
- **Report free-text** — user-controlled; could contain PII the user types themselves
  - **Potentional solution**: Periodic cleanup via time-expiration window.

<details>
  <summary>Affected Files</summary>

| File | Change |
|---|---|
| `interfaces/user.interface.ts` | Added `UserDataExport` type |
| `users.service.ts` | Added `exportUserData()` — single `findUnique` with nested includes for all four related tables, then an audit write |
| `users.controller.ts` | Added `GET me/data` handler with `ForbiddenException` guard on missing email |
| `users.service.spec.ts` | 3 new tests: happy path shape, no raw email in audit row, 404 on missing user |
| `users.controller.spec.ts` | 2 new tests: happy path response, 403 on missing email |
| `test/users.e2e-spec.ts` | New `me/data` describe block with disposable user; `MockAuthGuard` updated to route `/me/data` to a dedicated test email so the full stack is exercised end-to-end |
</details>


chuckmilton:
Resolved merge conflicts with `main` (commit `0c43924`).

**Conflicts touched:**
- `apps/backend/src/users/users.controller.ts`
- `apps/backend/src/users/users.controller.spec.ts`
- `apps/backend/src/users/users.service.ts`

**Resolution:** kept this PR's `GET /me/data` + `exportUserData()` alongside PR #174's `GET /me/forecast` + `getForecast()`. Both `me/*` routes are declared **before** `:userId`, so neither is shadowed by the IDOR-protected catch-all. Mock service in the spec carries both `exportUserData` and `getForecast` jest fns.

**Verified:**
- `tsc --noEmit` clean (backend tsconfig.build)
- `eslint .` clean
- 63/63 jest suites, 764/764 tests pass
- Prisma schema already has `USER_DATA_EXPORTED` enum value (migration `20260426221043_add_audit_events`) and the `User.reports` / `User.push_tokens` / `User.notification_logs` relations the export reads from
- e2e at `apps/backend/test/users.e2e-spec.ts` covers both shape (`profile.id` not exposed, `lot_id` is human-readable, all four arrays present) and audit row (`actor_hash` matches `^[a-f0-9]{64}$`, response body never contains the email)

**Reviewed:**
- `actor_hash` uses `SHA-256(DEVICE_HASH_SALT:email)` — same pattern as `USER_DELETED`, no reversible PII
- Internal `id` omitted from profile (covered by e2e assertion)
- All lot refs use `lot.lot_id` (human-readable like "G1"), not internal cuids
- No `:userId` param → no IDOR surface; caller identity is from JWT only
- 403 on missing JWT email (mirrors `deleteMe`)

**Follow-ups Zach flagged (out of scope, file as separate issues post-merge):**
- `notification_logs` retention prune cron
- `contributor_pings` retention cron
- `User.phone` column unused — drop in a future migration
- `Report.message` free-text PII risk — needs a length cap or moderation pass

Once CI is green this is ready to merge.

chuckmilton:
## Self-review: 4 issues found, 3 fixed in `3f73a01`

After resolving the merge conflict I went back through the diff with a security/correctness lens. Findings:

### 1. `GET /me/data` had no rate limit — **HIGH, fixed**
The endpoint reads the entire user record (favorites + reports + notification_logs + push_tokens) so it's both expensive and a high-value target for credential-stuffing / scraping. Added `@Throttle({ default: { limit: 3, ttl: 3_600_000 } })` (3/hour). ICO guidance for GDPR Art. 15/20 explicitly permits refusing "manifestly unfounded or excessive" repeat requests, so this ceiling is well within the law and well above any legitimate need.

### 2. Raw push tokens leaked into the export — **HIGH, fixed**
`push_tokens[].token` previously returned the full FCM/APNs value. Anyone holding a leaked export could send unauthenticated push notifications until the OS rotates the token. Now returns `{ token_preview, platform, registered_at }` where `token_preview` is the last 6 chars with a leading `…` — enough for the user to identify which device a token belongs to, useless for impersonation. Updated interface, e2e test now asserts the raw token never appears in the response body.

### 3. Audit row + read were not transactional — **MEDIUM, fixed**
Previously the audit was written first and the read happened in a separate query; a 404 (or any DB blip on the read) would leave a `USER_DATA_EXPORTED` row with no actual export, polluting the audit log. Now wrapped in an interactive `prisma.$transaction(async (tx) => …)` mirroring the `deleteUser` pattern — throw inside the callback rolls back the audit row. Unit test updated to verify no audit write happens for a missing user.

### 4. Unbounded result sets — **LOW, flagged for follow-up**
`notification_logs`, `reports`, and `push_tokens` are queried with no `take` limit. Combined with the indefinite retention noted in `docs/privacy-data-inventory.md`, a long-lived account could OOM the worker on export. This needs a separate retention/pagination story (probably tied to the data-deletion runbook), so I left it out of this PR rather than paper over it here.

### Verification
- `tsc --noEmit` ✓ (build + spec configs)
- `eslint .` ✓
- `pnpm test` → 764/764 unit ✓
- `pnpm exec jest --config ./jest-e2e.js --testPathPatterns=users.e2e` → 18/18 ✓ (incl. new push-token-mask test)

chuckmilton (commented):
Follow-up — Issue 4 + all four review comments addressed in `f81e8ca`.

**Issue 4 — unbounded relation reads in `/me/data`:** added per-relation `take` + deterministic `orderBy` inside the `tx.user.findUnique` include block (favorites:100/added_at desc, push_tokens:50/created_at desc, reports:1000/created_at desc, notification_logs:5000/sent_at desc). Caps documented inline.

**Zach's follow-ups:**

1. **`notification_logs` retention** — `NotificationsService.pruneOldLogs(retentionDays = 90)` + new `prune-notification-logs` cron (daily 04:15 PT, `NOTIFICATION_LOG_RETENTION_DAYS` override). Safe because the dedup window in `wasRecentlyNotified` is bounded by an explicit `gte: since` filter, so older rows are pure history.
2. **`contributor_pings` retention** — `ContributorService.pruneIdlePings(idleDays = 180)` + new `prune-contributor-pings` cron (Mondays 05:30 PT, `CONTRIBUTOR_PING_RETENTION_DAYS` override). Both `last_seen_at` AND `granted_at` must be stale before deletion (won't prune a recently re-granted device that hasn't pinged yet). 180d chosen so a student who skips the May–Aug recess isn't re-onboarded in Fall.
3. **`User.phone` removal** — column was unused everywhere (no input, no output, no validation). Migration `20260505133548_drop_user_phone` plus 13 reference sites cleaned (interfaces, seeds, specs, mappers).
4. **`Report.message` hardening** — picked the proper-fix combo over a simple cap:
   - DTO: `@MaxLength(REPORT_MESSAGE_MAX_LENGTH = 500)` exported and applied.
   - Service: server-side trim + truncate (defense in depth) + profanity censor on write via `bad-words@^3.0.4` (chosen for stable CJS API and a curated default list — censors with `*` rather than rejecting, so legitimate reports about rude behavior still go through).
   - Retention: `ReportsService.pruneOldMessages(retentionDays = 90)` + new `prune-old-report-messages` cron (Sundays 04:45 PT, `REPORT_MESSAGE_RETENTION_DAYS` override). Row is **retained** (type + lot + timestamp are aggregate signals); only the `message` column is set to `NULL`.

Mechanics:
- `bad-words` v3 has no DefinitelyTyped support for its current API → minimal ambient `apps/backend/src/types/bad-words.d.ts` shim (Filter constructor + `clean` / `isProfane` / addWords / removeWords). `tsconfig.spec.json` updated to include `src/**/*.d.ts`. `.gitignore` updated with negate exception for `apps/backend/src/types/*.d.ts` so hand-authored shims aren't swept up by the generic build-artifact rule.
- All 3 new cron entries added to `CRON_MONITORS` and the lockstep test (`scheduler.module.spec.ts`) so the boot-time mismatch check stays accurate.
- New services wired into `SchedulerModule.providers` directly (instead of importing `AuthModule`/`ReportsModule`) to avoid pulling Passport + UsersModule into the scheduler boot path.
- `docs/privacy-data-inventory.md` rows updated for all four retention changes.

Verified locally: lint clean, build + spec typecheck clean, **784 unit tests pass**, **26 e2e tests pass** (`users.e2e`, `reports.e2e`) against local pg with the new migration applied.

PR #193, opened 2026-05-04
fix(android): explicit android dependency

zachfpadilla:
# Android `RNGetRandomValues()` Crash

## Summary

- Fixes Android crash on startup: `Invariant Violation: TurboModuleRegistry.getEnforcing(...): 'RNGetRandomValues' could not be found`
- `react-native-get-random-values` v2.0.0 dropped the Android autolinking metadata from its `package.json`, so the RN CLI's `react-native config` command no longer detected it as a native module — `RNGetRandomValuesPackage` was never compiled into the app
- Added an explicit `dependencies` entry in `react-native.config.js` to force the autolinking step to include the package, restoring `RNGetRandomValues` in the generated `PackageList.java`

## Root cause

`react-native-get-random-values` v2.0.0 removed the `"android"` autolinking field from its `package.json` (which older versions used to register `packageImportPath` / `packageInstance`). The RN CLI's autolinking scan relies on that field to discover native packages. Without it, the module is silently skipped during the Gradle codegen step, but the JS side still calls `TurboModuleRegistry.getEnforcing('RNGetRandomValues')` — crashing at startup before any lots can be fetched.

## Changes

- `apps/mobile/react-native.config.js` — added explicit `dependencies['react-native-get-random-values'].platforms.android` config so the autolinking generates `new org.linusu.RNGetRandomValuesPackage()` in `PackageList.java`

## Notes

- You must clean the Android build before running:
`cd apps/mobile/android && ./gradlew clean`


DottL (approved):
🌻 Thank you !!

chuckmilton (approved):

PR #197, opened 2026-05-05
feat: wire push notifications to mobile

lawrenceDegoma:
## Summary

Wires Firebase Cloud Messaging (FCM) push notifications end-to-end on mobile, connecting the backend notification infrastructure to the React Native app.

---

### New Files
- **`src/services/pushNotifications.ts`** — FCM lifecycle service: requests permission (iOS), retrieves the FCM token, registers it with the backend, sets up a token-refresh listener, and exposes foreground/background message subscription helpers.
- **`src/services/api/notifications.ts`** — API wrapper for `POST /users/me/push-token`; sends the device token and platform (`ios`/`android`) to the backend with Bearer auth.
- **`__mocks__/@react-native-firebase/app.js`** — Jest manual mock so tests never touch the native Firebase bridge.
- **`__mocks__/@react-native-firebase/messaging.js`** — Jest manual mock that stubs the full FCM API surface used by our service.

### Modified Files
- **`src/context/AuthContext.tsx`**
  - Calls `initPushNotifications()` after login and on app init (when a saved session is restored); calls `stopPushNotifications()` on logout.
  - Decodes `idToken` with `jwtDecode` to expose `userEmail` (used by the notification preferences API).
- **`App.tsx`**
  - Registers a foreground message handler (shows an `Alert` with title + body).
  - Registers a background tap handler that deep-links the user to the relevant screen (Map → Short Term Forecast for lot notifications; Map root for surge/event alerts).
  - Checks `getInitialNotification()` on cold start so taps from a quit state also navigate correctly.
- **`src/screens/ProfileScreen.tsx`**
  - Replaces the old placeholder toggles with four real preference toggles (`Favorite Lot Filling`, `Favorite Lot Clearing`, `Parking Surge Alerts`, `Campus Event Alerts`).
  - Loads current preferences from the backend on mount and calls `PATCH /users/:email/notifications` on each toggle with optimistic UI + rollback on error.
- **`src/services/api/users.ts`** — Added `updateNotificationPreferences()`.
- **`src/services/api/index.ts`** — Re-exports `registerPushToken` and `updateNotificationPreferences`.
- **`jest.config.js`** — Added `moduleNameMapper` entries to redirect `@react-native-firebase/*` imports to the manual mocks.
- **`__tests__/useFavorites.test.ts`** — Added `userEmail` field to both mock `useAuth` return values to satisfy the updated `AuthContextType`.

---

## Testing

- All 689 existing tests pass (`57` suites).
- Firebase native modules are fully mocked — no native bridge required for Jest.
- `ForceUpdateScreen.test.tsx` is an untracked file from a separate branch and is **not** part of this PR.

---

## Dependencies Added

| Package | Reason |
|---|---|
| `@react-native-firebase/app` | Firebase core (required peer dep) |
| `@react-native-firebase/messaging` | FCM token retrieval & message handling |

iOS pods updated (`Podfile.lock`).


lawrenceDegoma:
Ngl i'm not sure why the ci checks are failing. I see they're a backend thing but I don't think that's anything from me 
image

chuckmilton:
## Review — push notifications wiring

Did a thorough end-to-end pass against `main`, the backend, and native iOS/Android setup. Resolved the merge conflict in `App.tsx` additively (push handlers + `useForceUpdate` coexist) and pushed two commits — one for the merge, one for the issues below that I went ahead and fixed.

### Pushed fixes (commit `166db1f`)

1. **CRITICAL — wrong HTTP verb on notification preferences.** `apps/mobile/src/services/api/users.ts` called `apiService.put` against `/users/:userId/notifications`, but the backend route is `@Patch(':userId/notifications')` (`apps/backend/src/users/users.controller.ts:139`). Every toggle would have hit a 404/405 in production. The JSDoc on the function even said “PATCH”. Added an `apiService.patch` helper to `base.ts` and switched the call.

2. **HIGH — toggles never hydrated from the backend.** The PR description says “Loads current preferences from the backend on mount”, but `ProfileScreen` just initialised state with hardcoded defaults (`{filling:true, clearing:true, surge:false, events:true}`) and only ever wrote on toggle. So a user with stored preferences would see the wrong switch positions until they tapped each one (then optimistic-update would overwrite the server with the wrong value). Added `getUserProfile()` calling `GET /users/:userId` and a `useEffect` in `ProfileScreen` that hydrates `notifPrefs` from `user.notification_preferences` when `userEmail` is available. Falls back to local defaults on auth/network failure.

3. **HIGH — iOS `pod install` was broken.** Fresh install fails with:
   > The Swift pod `FirebaseCoreInternal` depends upon `GoogleUtilities`, which does not define modules.

   Pinned `GoogleUtilities` and the relevant Firebase pods to `:modular_headers => true` in `ios/Podfile` (matching the existing `AppAuth` workaround). Without this nobody can build iOS off this branch.

### Things I left alone but you should know about

4. **HIGH — Android Firebase setup is missing entirely.** No `google-services.json` under `android/app/`, no `classpath 'com.google.gms:google-services'` in `android/build.gradle`, no `apply plugin: 'com.google.gms.google-services'` in `android/app/build.gradle`. As shipped, push notifications **will silently no-op on Android** — `messaging().getToken()` will throw or return null and the catch-all in `initPushNotifications` swallows it. Needs a Firebase Console project with an Android app entry, the `google-services.json` downloaded into `android/app/`, and the gradle plugin wired up. I didn't generate this because it requires Firebase Console access and a real package-name registration.

5. **MEDIUM — push token never unregistered server-side on logout.** `stopPushNotifications` only tears down the local `onTokenRefresh` listener; the backend still has the FCM token associated with the previously-signed-in account. If user A logs out and user B signs in on the same physical device, B's login *will* re-register the same token under B (so notifications won't leak), but the row under A remains and will keep getting targeted by any "send to all of A's devices" path until the token rotates or A's account is purged. Consider a `DELETE /users/me/push-token` and calling it from `stopPushNotifications`. Out of scope for this PR but worth a follow-up issue.

6. **LOW — `setTimeout(…, 500)` for `getInitialNotification`.** Functional but fragile; race condition on slow cold starts. `NavigationContainer` exposes an `onReady` callback that's the canonical place to hook this. Non-blocking.

7. **Tiny nit — `__mocks__/@react-native-firebase/app.js`** has `module.exports = {…}` written twice back to back. Harmless (second overwrites the first with an identical value) but makes a reviewer do a double-take. Worth deleting one of them next time you touch the file.

### What I verified

- ✅ `tsc --noEmit` clean across the mobile workspace post-merge + post-fixes.
- ✅ `pnpm test` — **59 suites / 696 tests pass**, no regressions from the merge or the fixes.
- ✅ Backend `azure.strategy.ts` accepts both `id_token` and `access_token` audiences, so the token-type inconsistency between `notifications.ts` (uses `accessToken`) and `users.ts` (uses `idToken`) is not a bug — but it's worth standardising on one for consistency.
- ✅ Navigation routing in `handleNotificationNavigation` lines up with the existing `MapStackParamList` (`'Short Term Forecast'` route exists in `MainTabNavigator`).
- ✅ Backend `notification_preferences` schema (Prisma JSON) and the merge logic in `users.service.ts:140-160` match the mobile DTO shape (`favorites_filling | favorites_clearing | surge_alerts | event_alerts`).
- ✅ iOS `AppDelegate.swift` has `FirebaseApp.configure()` and `GoogleService-Info.plist` is committed.
- ✅ Conflict resolution + auto-merged `pnpm-lock.yaml` clean; `git diff --check` reports no whitespace damage.

### Suggested follow-ups (not in this PR)

- File an issue: Android Firebase setup (item 4) — blocker before we can claim push works on Android.
- File an issue: backend `DELETE /users/me/push-token` + wire from logout (item 5).
- Consider replacing `setTimeout(500)` with `NavigationContainer.onReady` (item 6).

chuckmilton:
Follow-up commit b8cbdf2 resolves the three remaining items from my prior review:

**4. HIGH — Android Firebase setup** ✅
- Added `com.google.gms:google-services:4.4.2` classpath in `android/build.gradle`
- Applied `com.google.gms.google-services` plugin in `android/app/build.gradle` (build will fail loudly if config is missing — no silent no-op)
- Committed `android/app/google-services.json` (project: sharkpark-dev, package: app.sharkpark.mobile — matches applicationId). Note: the `current_key` is a client API key (ships in every APK); per [Firebase docs](https://firebase.google.com/docs/projects/api-keys) it's safe to commit. It is restricted by Android package + SHA-1 in GCP Console.

**5. MEDIUM — Push token unregistered on logout** ✅
- Backend: new `DELETE /users/me/push-token` endpoint with `UnregisterPushTokenDto { token: string }`. Service uses `prisma.pushToken.deleteMany({ where: { token, user_id: user.id } })` — scoped by `user_id` so a hostile actor with a leaked FCM token cannot evict another user's tokens. Idempotent (no-op if user/token already gone).
- Mobile: `unregisterCurrentPushToken()` in `pushNotifications.ts` reads the FCM token and calls the API. Invoked from `AuthContext.logout()` **before** `logoutFromAzure` so the bearer token is still available.
- Tests: 2 new controller specs + 3 new service specs + DTO validation spec.

**6. LOW — `setTimeout(500)` for `getInitialNotification`** ✅
- Replaced the timer with a module-level `pendingInitialNotificationData` bridge that drains via `NavigationContainer`'s `onReady` callback (or immediately if nav is already ready when the FCM payload resolves). No more arbitrary 500ms guess.

Verification: `tsc --noEmit` clean for both packages; 792 backend tests pass; 696 mobile tests pass; eslint clean on touched files.

PR #203, opened 2026-05-05
fix(mobile): navigate to exact lot coordinates in Apple Maps

lawrenceDegoma:
## Problem
Tapping "Navigate" for any parking lot and selecting Apple Maps redirected to **California State University - Long Beach** (the campus) instead of the specific lot.

**Root cause:** `react-native-map-link` builds the Apple Maps URL as: maps://?daddr=33.78,-118.11&q=Parking+Lot+G7&dirflg=d

When both `daddr=` and `q=` are present, Apple Maps ignores the coordinates entirely and performs a **text search** on `q=`, resolving "Parking Lot G7" to the nearest named landmark — the campus.

---

## Solution
Override the `open()` function for the `apple-maps` entry returned by `getApps()` to use the correct URL format: maps://?ll=33.78,-118.11&q=Parking+Lot+G7&dirflg=d

With `ll=` present, Apple Maps **anchors the pin to the exact coordinates** and treats `q=` purely as a display label — no text search occurs. Driving directions open directly to the lot.

Also added `maps` to `LSApplicationQueriesSchemes` in `Info.plist`, which is required for iOS to resolve the `maps://` scheme.

---

## Changes
| File | Change |
|---|---|
| `src/components/Modals/MapSelectModal.tsx` | Override `apple-maps` `open()` to use `ll=`-anchored URL via `Linking.openURL` |
| `ios/mobile/Info.plist` | Add `maps` to `LSApplicationQueriesSchemes` |
| `__tests__/MapSelectModal.test.tsx` | Update interaction test to assert `Linking.openURL` is called with the coordinate-anchored URL |

---

## Testing
- **Unit tests:** All 7 `MapSelectModal` tests pass (696 total, 0 failures)
- **Manual:** Rebuilt native app, tapped Apple Maps for multiple lots — each opens a named pin dropped at the lot's exact GPS coordinates with driving directions

---

## URL Format Reference
| Format | Apple Maps behavior |
|---|---|
| `daddr=lat,lon&q=<name>`  | Text-searches `<name>`, ignores coordinates |
| `ll=lat,lon&q=<name>`  | Drops pin at exact `lat,lon`, uses `<name>` as label

### DIFF
--- a/apps/mobile/__tests__/MapSelectModal.test.tsx
+++ b/apps/mobile/__tests__/MapSelectModal.test.tsx
@@ -9,6 +9,7 @@
  */
 import React from 'react';
 import ReactTestRenderer from 'react-test-renderer';
+import { Linking } from 'react-native';
 import { MapSelectModal } from '../src/components/Modals/MapSelectModal';
 import { getApps } from 'react-native-map-link';
 
@@ -155,6 +156,8 @@ describe('MapSelectModal', () => {
 
   describe('interactions', () => {
     it('calls open() on the selected app and triggers close animation', async () => {
+      const openURLSpy = jest.spyOn(Linking, 'openURL').mockResolvedValue(undefined);
+
       let tree: ReactTestRenderer.ReactTestRenderer;
       await ReactTestRenderer.act(async () => {
         tree = ReactTestRenderer.create(<MapSelectModal {...defaultProps} />);
@@ -168,7 +171,11 @@ describe('MapSelectModal', () => {
         await touchable!.props.onPress();
       });
 
-      expect(mockAppOpen1).toHaveBeenCalledTimes(1);
+      // Apple Maps open() is overridden to use Linking.openURL with coordinate-anchored URL
+      expect(openURLSpy).toHaveBeenCalledTimes(1);
+      expect(openURLSpy).toHaveBeenCalledWith(
+        `maps://?ll=${defaultProps.lat},${defaultProps.lon}&q=${encodeURIComponent(defaultProps.title)}&dirflg=d`
+      );
       expect(mockAppOpen2).not.toHaveBeenCalled();
 
       // Animation completion
@@ -177,6 +184,7 @@ describe('MapSelectModal', () => {
       });
 
       expect(defaultProps.onClose).toHaveBeenCalledTimes(1);
+      openURLSpy.mockRestore();
     });
 
     it('calls onClose when the close icon is pressed', async () => {
--- a/apps/mobile/ios/mobile/Info.plist
+++ b/apps/mobile/ios/mobile/Info.plist
@@ -45,6 +45,16 @@
 	</array>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<!--
+	  Export-compliance declaration. SharkPark only uses HTTPS/TLS via Apple's
+	  Network framework + standard URLSession, which qualifies for the
+	  exemption under EAR 740.17(b)(1) (mass-market software using only
+	  encryption already provided by the operating system). Setting this to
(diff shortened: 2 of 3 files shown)


chuckmilton (commented):
Thanks for the deep dive — the URL diagnosis (`daddr=` + `q=` → text search → campus) is correct and the fix is the right shape. A few changes needed before merge.

### 🔴 Blocking

**1. `apps/mobile/ios/Info.plist` — restore the export-compliance comment**

This PR removes a 9-line comment block explaining why `ITSAppUsesNonExemptEncryption=false` is set. That comment exists for a legal reason (EAR 740.17(b)(1) exemption) and is unrelated to the Apple Maps fix. Adding `<string>maps</string>` to `LSApplicationQueriesSchemes` is the only change needed in this file.

Please restore the comment block exactly as it was on `main`.

**2. `MapSelectModal.tsx` — stale outer comment contradicts the code**

```ts
// Apple Maps resolves `q=<lot name>` to the nearest named landmark (e.g. the campus)
// instead of navigating to the exact coordinates. Override the open() function for
// apple-maps to use `daddr=lat,lon` only — no text query parameter.
```

The actual implementation is `ll=lat,lon&q=<name>&dirflg=d` — it does include `q=`, just with `ll=` to anchor the pin. Either delete the outer comment (the inner one already explains it correctly) or rewrite it to match. As-is it'll mislead the next reader.

### 🟡 Nit

**3. `MapSelectModal.tsx` — wrap `Linking.openURL` in try/catch**

```ts
open: async () => {
  const encodedTitle = encodeURIComponent(title);
  const url = `maps://?ll=${lat},${lon}&q=${encodedTitle}&dirflg=d`;
  await Linking.openURL(url);
},
```

`Linking.openURL` rejects if the scheme can't be opened (e.g. the user has somehow removed Apple Maps, or a future iOS sandboxing change blocks `maps://`). Right now that rejection bubbles up as an unhandled promise rejection because the caller in `FlatList`'s `onPress` doesn't await it. Match the pattern used elsewhere in this file (`Alert.alert(...)`) so the user gets a graceful error rather than the modal closing on a silent failure.

```ts
open: async () => {
  const encodedTitle = encodeURIComponent(title);
  const url = `maps://?ll=${lat},${lon}&q=${encodedTitle}&dirflg=d`;
  try {
    await Linking.openURL(url);
  } catch (err) {
    console.error('Failed to open Apple Maps:', err);
    Alert.alert('Could not open Apple Maps', 'Please try a different navigation app.');
  }
},
```

### ℹ️ CI status (not yours to fix)

The `ios-build` failure is the FirebaseCore/FirebaseInstallations modulemap-redefinition issue introduced by Xcode 16's explicit-modules default — already fixed in #205. Once that lands, rebase and CI should go green.

### ✅ Looks good
- URL format change is correct; `ll=` + `q=` + `dirflg=d` is the documented Apple Maps pattern for "drop a pin at exact coordinates with a label and start driving directions"
- `encodeURIComponent(title)` properly handles spaces / special chars in lot names
- Test assertion matches the new URL exactly, including encoded title
- `LSApplicationQueriesSchemes` `maps` entry is required for `Linking.canOpenURL('maps://')` to return true on iOS 9+ — necessary even though we use `openURL` directly, because some iOS versions still gate this

chuckmilton:
Pushed fixes addressing all three review points (commit on Feat/lot-maps-navigation):

1. ✅ Restored EAR 740.17(b)(1) export-compliance comment in Info.plist
2. ✅ Rewrote stale outer comment in MapSelectModal to match the actual ll= + q= + dirflg=d implementation
3. ✅ Wrapped Linking.openURL in try/catch + Alert (matches existing error-handling pattern in this file)

All 7 MapSelectModal tests pass locally. Once #205 merges (FirebaseCore modulemap fix), please rebase to clear the ios-build red.

PR #207, opened 2026-05-06
fix(mobile): close report modal and stay on lot after submission

lawrenceDegoma:
# fix(mobile): close report modal and stay on lot after submission

## Summary

This PR fixes two issues:

1. **Report modal navigation bug** — submitting a report (or attempting to while unauthenticated) was redirecting the user to the Profile tab instead of keeping them on the lot screen.
2. **iOS build failure** — Xcode 26 introduced a module system change that, combined with a duplicate Firebase integration (CocoaPods + SPM), caused the build to fail with `redefinition of module 'FirebaseCore'`.

---

### What was wrong

In `ShortTermForecastScreen.tsx`, `handleReportSubmit` had two calls to `navigation.navigate('Profile' as never)`:

```tsx
// Case 1: user is a guest or not authenticated
if (isGuest || !isAuthenticated) {
  setIsReportModalOpen(false);
  navigation.navigate('Profile' as never); //  wrong — sends user to Profile tab
  return;
}

// Case 2: server returns 401/403
} catch (err) {
  if (err instanceof ReportUnauthorizedError) {
    setIsReportModalOpen(false);
    navigation.navigate('Profile' as never); //  wrong — same problem
    return;
  }
}
```

Both cases were supposed to prompt the user to sign in, but instead silently dropped them into the Profile tab with no explanation. Additionally, a **successful** submission never called `setIsReportModalOpen(false)`, so the modal remained open after the report was sent.

### What we changed

- **Guest / unauthenticated:** Close the modal and show `Alert.alert('Sign in required', 'Please sign in to submit a report.')` instead of navigating away.
- **`ReportUnauthorizedError`:** Same — close modal, show alert.
- **Successful submission:** Added `setIsReportModalOpen(false)` so the modal actually closes.

```tsx
// After
if (isGuest || !isAuthenticated) {
  setIsReportModalOpen(false);
  Alert.alert('Sign in required', 'Please sign in to submit a report.'); // 
  return;
}
// ...
await reportsApi.create({ ... });
setIsReportModalOpen(false); //  close on success
// ...
if (err instanceof ReportUnauthorizedError) {
  setIsReportModalOpen(false);
  Alert.alert('Sign in required', 'Please sign in to submit a report.'); // 
  return;
}
```

**Files changed:** `apps/mobile/src/screens/ShortTermForecastScreen.tsx`

---

## 2. iOS Build Failure — Root Cause Analysis & Fix

This took significant investigation because the failure had three distinct, compounding layers. Here is the full chain.

### Environment

| Component | Version |
|---|---|
| Xcode | 26.0 (Build 17A324) |
| iOS Simulator SDK | 26.0 |
| CocoaPods (before fix) | 1.15.2 |
| CocoaPods (after fix) | 1.16.2 |
| Ruby | 3.3.11 (via rbenv) |
| `@react-native-firebase/app` | 24.0.0 |
| Firebase iOS SDK (CocoaPods) | 12.10.0 |

---

### Layer 1 — Xcode 26 Explicitly Built Modules

**Error:**
```
Pods/Headers/Public/FirebaseCore/FirebaseCore.modulemap:1:8:
  error: redefinition of module 'FirebaseCore'
```

**Root cause:**

Xcode 26 introduced **Explicitly Built Modules** (`CLANG_ENABLE_EXPLICIT_MODULES`), a new compilation model where Xcode auto-generates its own module map for every target it builds — regardless of whether the target already ships a module map of its own.

Our `Podfile` had the following lines, which are required for `@react-native-firebase` to compile because Firebase's Swift pods depend on GoogleUtilities:

```ruby
pod 'GoogleUtilities',        :modular_headers => true
pod 'FirebaseCore',           :modular_headers => true
pod 'FirebaseCoreInternal',   :modular_headers => true
pod 'FirebaseInstallations',  :modular_headers => true
```

`:modular_headers => true` tells CocoaPods to generate and expose a module map for these pods. Xcode 26 then *also* generated its own module map for the same pods. The compiler saw two module definitions for `FirebaseCore` (one from CocoaPods, one from Xcode 26's auto-generation) and failed.

**Fix:**

Disable `CLANG_ENABLE_EXPLICIT_MODULES` on both the Pods project and the main app project via `post_install`:

```ruby
[installer.pods_project, installer.aggregate_targets.first&.user_project].compact.each do |proj|
  proj.build_configurations.each do |cfg|
    cfg.build_settings['CLANG_ENABLE_EXPLICIT_MODULES'] = 'NO'
  end
  proj.targets.each do |target|
    target.build_configurations.each do |cfg|
      cfg.build_settings['CLANG_ENABLE_EXPLICIT_MODULES'] = 'NO'
    end
  end
  proj.save
end
```

We also added `DEFINES_MODULE = NO` as a belt-and-suspenders measure for the specific Firebase pods that already supply their own module maps, to prevent CocoaPods itself from re-emitting them:

```ruby
firebase_pods_with_own_modulemaps = %w[
  FirebaseCore FirebaseCoreInternal FirebaseCoreExtension
  FirebaseInstallations FirebaseMessaging GoogleUtilities
]
installer.pods_project.targets.each do |target|
  next unless firebase_pods_with_own_modulemaps.include?(target.name)
  target.build_configurations.each do |cfg|
    cfg.build_settings['DEFINES_MODULE'] = 'NO'
  end
end
```

---

### Layer 2 — CocoaPods 1.15.2 Double-Emitting `-fmodule-map-file=` Flags

**Root cause:**

CocoaPods 1.15.2 had a known bug where it emitted duplicate `-fmodule-map-file=` compiler flags for certain pods in the generated xcconfig files. Under Xcode 26's stricter module validation, duplicate flags triggered the same `redefinition` error through a second code path.

**Fix:** Upgraded CocoaPods from **1.15.2 → 1.16.2** using Ruby 3.3.11 via rbenv. CocoaPods 1.16.x deduplicated the flag emission, eliminating this second trigger.

---

### Layer 3 (Root Cause) — Duplicate Firebase Integration: CocoaPods + Direct SPM Reference

**Error (after Layer 2 fix):**
```
invalid custom path 'FirebaseSharedSwift/Sources' for target 'FirebaseSharedSwift'
```

**Root cause:**

At some point in the project's history, a direct **Swift Package Manager (SPM) reference** to `firebase-ios-sdk` was added to `mobile.xcodeproj/project.pbxproj`:

```
81A4221E2FA47A2D004C53DF /* XCRemoteSwiftPackageReference "firebase-ios-sdk" */ = {
    isa = XCRemoteSwiftPackageReference;
    repositoryURL = "https://github.com/firebase/firebase-ios-sdk";
    requirement = {
        kind = upToNextMajorVersion;
        minimumVersion = 12.12.1;   ← this is the problem
    };
};
```

This means Firebase was being pulled in **twice**:

| Integration | Version resolved |
|---|---|
| CocoaPods (via `react-native-firebase@24.0.0`) | **12.10.0** (pinned by RNFirebase) |
| Direct SPM reference in `project.pbxproj` | **12.13.0** (resolved from `upToNextMajorVersion 12.12.1`) |

Xcode independently resolved the SPM reference to the latest compatible version (12.13.0) while CocoaPods stayed on 12.10.0. The version mismatch caused the `FirebaseSharedSwift/Sources` path error because 12.13.0 restructured that module's source layout.

**Investigation path:**

Pinning `$FirebaseSDKVersion = '12.10.0'` in the `Podfile` only controls CocoaPods — it does not affect Xcode's independent SPM resolution. Changing the SPM requirement to `exactVersion = 12.10.0` also didn't fully resolve it because Firebase was still being loaded twice regardless of version alignment: CocoaPods provided the module maps via `:modular_headers`, and SPM provided a second parallel copy of the same frameworks, giving the compiler two competing module definitions.

**Fix:**

Removed the direct SPM `firebase-ios-sdk` reference from `project.pbxproj` entirely — all four associated entries:

1. `PBXBuildFile` for `FirebaseMessaging in Frameworks`
2. The `FirebaseMessaging` entry in `PBXFrameworksBuildPhase`
3. The `XCRemoteSwiftPackageReference` block for `firebase-ios-sdk`
4. The `XCSwiftPackageProductDependency` block for `FirebaseMessaging`

CocoaPods is now the **sole** integration point for Firebase. The `@react-native-firebase/messaging` pod already pulls in `FirebaseMessaging` as a transitive dependency via CocoaPods, so the SPM reference was entirely redundant.

---

### Why All Three Layers Had to Be Fixed

Fixing Layer 1 alone (disabling `CLANG_ENABLE_EXPLICIT_MODULES`) suppressed the initial `redefinition` error but unmasked the SPM version mismatch error underneath. Fixing Layer 2 (CocoaPods upgrade) was necessary for stable xcconfig generation. Only after fixing Layer 3 (removing the duplicate SPM reference) did the build succeed, because it eliminated the duplicate module loading at the source.

---

## Files Changed

| File | Change |
|---|---|
| `apps/mobile/src/screens/ShortTermForecastScreen.tsx` | Replace `navigation.navigate('Profile')` with `Alert.alert`; close modal on success |
| `apps/mobile/ios/Podfile` | Add `$FirebaseSDKVersion = '12.10.0'`; add `CLANG_ENABLE_EXPLICIT_MODULES = NO` and `DEFINES_MODULE = NO` in `post_install` |
| `apps/mobile/ios/Podfile.lock` | Regenerated after `pod install` with CocoaPods 1.16.2 |
| `apps/mobile/ios/mobile.xcodeproj/project.pbxproj` | Remove `XCRemoteSwiftPackageReference` for `firebase-ios-sdk` and all associated entries |

## Testing

- All 706 existing Jest unit/component tests pass (`pnpm test`)
- `xcodebuild` against iPhone 17 simulator (iOS 26.0): **BUILD SUCCEEDED**
- Report modal: verified modal closes in-place on success, alert shown for unauthenticated users

### DIFF
--- a/apps/mobile/ios/Podfile
+++ b/apps/mobile/ios/Podfile
@@ -8,6 +8,12 @@ require Pod::Executable.execute_command('node', ['-p',
 platform :ios, min_ios_version_supported
 prepare_react_native_project!
 
+# Pin Firebase SDK version to match react-native-firebase@24.0.0 expectations.
+# Without this, CocoaPods 1.16.x SPM resolver independently picks up a newer
+# Firebase version (e.g. 12.13.0) while CocoaPods itself stays on 12.10.0,
+# causing 'invalid custom path FirebaseSharedSwift/Sources' at build time.
+$FirebaseSDKVersion = '12.10.0'
+
 linkage = ENV['USE_FRAMEWORKS']
 if linkage != nil
   Pod::UI.puts "Configuring Pod with #{linkage}ally linked Frameworks".green
@@ -51,6 +57,40 @@ target 'mobile' do
         target.build_configurations.each do |config|
           config.build_settings['CLANG_CXX_LANGUAGE_STANDARD'] = 'c++17'
         end
+      end
+    end
+
+    # Xcode 26 introduced "Explicitly Built Modules" which generates its own
+    # module maps for every pod regardless of DEFINES_MODULE, causing
+    # "redefinition of module 'FirebaseCore'" for pods that already ship module
+    # maps via :modular_headers => true. Disabling it on the Pods project,
+    # all pod targets, AND the main app project fixes the cascade.
+    [installer.pods_project, installer.aggregate_targets.first&.user_project].compact.each do |proj|
+      proj.build_configurations.each do |cfg|
+        cfg.build_settings['CLANG_ENABLE_EXPLICIT_MODULES'] = 'NO'
+      end
+      proj.targets.each do |target|
+        target.build_configurations.each do |cfg|
+          cfg.build_settings['CLANG_ENABLE_EXPLICIT_MODULES'] = 'NO'
+        end
+      end
+      proj.save
+    end
+
+    # Belt-and-suspenders: also keep DEFINES_MODULE off for Firebase pods that
+    # already supply their own module maps via :modular_headers above.
+    firebase_pods_with_own_modulemaps = %w[
+      FirebaseCore
+      FirebaseCoreInternal
+      FirebaseCoreExtension
+      FirebaseInstallations
+      FirebaseMessaging
+      GoogleUtilities
+    ]
+    installer.pods_project.targets.each do |target|
+      next unless firebase_pods_with_own_modulemaps.include?(target.name)
+      target.build_configurations.each do |cfg|
+        cfg.build_settings['DEFINES_MODULE'] = 'NO'
       end
     end
     # ----------------------------
--- a/apps/mobile/ios/Podfile
+++ b/apps/mobile/ios/Podfile
@@ -65,7 +65,8 @@ target 'mobile' do
     # "redefinition of module 'FirebaseCore'" for pods that already ship module
     # maps via :modular_headers => true. Disabling it on the Pods project,
     # all pod targets, AND the main app project fixes the cascade.
-    [installer.pods_project, installer.aggregate_targets.first&.user_project].compact.each do |proj|
+    user_projects = installer.aggregate_targets.map(&:user_project).compact.uniq
+    [installer.pods_project, *user_projects].each do |proj|
       proj.build_configurations.each do |cfg|
         cfg.build_settings['CLANG_ENABLE_EXPLICIT_MODULES'] = 'NO'
       end
--- a/apps/mobile/ios/Podfile
+++ b/apps/mobile/ios/Podfile
@@ -8,10 +8,16 @@ require Pod::Executable.execute_command('node', ['-p',
 platform :ios, min_ios_version_supported
 prepare_react_native_project!
 
-# Pin Firebase SDK version to match react-native-firebase@24.0.0 expectations.
-# Without this, CocoaPods 1.16.x SPM resolver independently picks up a newer
+# Pin Firebase SDK version to match @react-native-firebase/app@24.x expectations.
+# Without this, CocoaPods 1.16.x's SPM resolver independently picks up a newer
 # Firebase version (e.g. 12.13.0) while CocoaPods itself stays on 12.10.0,
 # causing 'invalid custom path FirebaseSharedSwift/Sources' at build time.
+#
+# BUMP CONTRACT: when upgrading @react-native-firebase/* major versions, update
+# this pin to match the Firebase iOS SDK version listed in that release's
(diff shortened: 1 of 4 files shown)


lawrenceDegoma:
# Added post all the headache fixing:
**THE GIRL IS GONE**
- MaterialIcons font wasn't resolving so its unicode fell back to the system emoji font which was the girl

## Fix:
- Swap to Ionicons which was already being imported anyways

zachfpadilla (approved):
Image

Issue #141, opened 2026-04-30
[Future] Map & Shuttle UI Changes


zachfpadilla:
# Map & Shuttle UI Changes
General issue for UI/UX changes regarding the map & map markers.

## Map Changes
- [x] Replace lot icons with lot-fitted polygons
   - Should be doable by binding a border to the defined Geofence perimeters for each lot.
- [x] Add shuttle/stop/route options to `FilterModal`
   - Currently, all transit data is rendered by default. Making certain aspects toggleable could allow for a cleaner interfaces if the user desires it (maybe revisit after deprecating lot icons).

## Shuttle Changes
- [x] Replace shuttle callout w/ a modal or bottom sheet
   - `react-native-maps` custom callouts are very finicky, especially in light of the new live data stream.
- [x] Make shuttle marker arrow orientations independent of map rotation
   - `flat` prop isn't being respected for the `ShuttleMarker` custom marker. Either correct orientation based off of current map angle, or find `flat` fix/override.
- [x] **(Potential)** Correct stuttering shuttle positions via linear interpolation
   - Despite the new websocket, shuttles still don't appear to travel smoothly (likely due to a combination of websocket speed & network latency). This is how transit options tend to appear in apps such as Apple/Google Maps, but could still potentially be accounted for via animating the shuttle's coordinate change over the course of its given velocity.

PR #210, opened 2026-05-06
feat: map and shuttle ui fixes

zachfpadilla:
# Map & Shuttle UI Fixes

## Summary

- Replaces lot circle markers with interactive polygons overlays on the map
- Adds multi-colored stop markers (one segment per route) via new `SegmentedCircle` component
- Adds a Nearby Transit card showing live ETAs for stops near the selected lot
- Adds a Transit tab to the filter modal with per-route toggles, persisted via `AsyncStorage`
- Addresses issues & features mentioned in: https://github.com/SharkPark-App/SharkPark/issues/141
- Fixes live/static shuttle data desyncing on WebSocket reconnect or new bus appearance
- Fixes phantom 0-minute ETAs caused by PassioGO's `"arrived"` sentinel being parsed as numeric

## Changes

**Backend**
- `PassioWebSocketService`: added 200 ms update batching, 30 s ping keep-alive, proper `terminate()` on error, and full `disconnect()` cleanup
- `ShuttleTrackerService`: added stale-bus pruning (30 s interval, 2 min TTL), `applyLiveUpdates()`, position-preserving merge in `fetchShuttles()` and `syncFromRedis()`, `metadataFetchPending` guard
- ETA parsing: skips `"arrived"` sentinel, prefers `etaR` over regex-parsed `eta` string
- `MapStop.routeId` → `routeIds: string[]` to support multi-route stops

**Frontend**
- New `SegmentedCircle` SVG component for multi-route stop markers
- New `NearbyTransitCard` + `useNearbyStopETAs` hook (15 s poll, no spinner on background refresh)
- New `transitProximity` utils: `nearbyStopsForLot`, `groupArrivals`, `formatEtas`
- New `lotPolygons` data for map overlays and proximity calculations
- `ShuttleMarker`: animated positions tween over 8.5 s; heading corrected for map bearing
- `MapScreen`: polygon lot overlays, segmented stop markers, transit filter state with `AsyncStorage` persistence
- `FilterModal`: added Transit tab with per-route toggles and tab-aware "Toggle all"
- `StopModal`: arrivals use shared `groupArrivals`/`formatEtas`; section is now scrollable
- `useTransitData`: functional updater preserves live positions on refresh; 60 s pruning poll; unknown bus IDs silently dropped

## Screenshots
| Main Map | Transit Filter | Nearby Stops |
|:-:|:-:|:-:|
| IMG_6030 | IMG_6031 | IMG_6032 |

## Tests
- 50 new tests across 5 new suites (`SegmentedCircle`, `ShuttleModal`, `NearbyTransitCard`, `useNearbyStopETAs`, `transitProximity`); 7 existing suites updated


chuckmilton (commented):
## Review notes

Solid PR — batching, segmented stops, lot polygons, persisted filters, and the daily-cron-driven metadata refresh all hang together well. I pushed one fix-up commit (c883949) for two latent backend bugs in `shuttle-tracker.service.ts`; the rest below are items worth considering before merge.

### Pushed in c883949 (fix-up commit)

1. **`applyLiveUpdates` dropped first WS frame for unknown buses.** The previous version recorded `lastSeen` for unknown bus IDs but only patched positions for buses already in `latestShuttles` (via `latestShuttles.map(...)`), so the very first WS frame's lat/lng was silently discarded. The bus didn't appear to clients until the on-demand `fetchShuttles()` returned (~REST round-trip + 60s mobile poll cadence). Fix: insert a placeholder `MapShuttle` for unknown IDs so the position is preserved; metadata is backfilled when `fetchShuttles()` lands (live position is preserved across that overwrite via the existing `livePositions` map).
2. **`applyLiveUpdates` was O(N×M).** Inner `updates.find(u => u.id === shuttle.id)` runs per existing shuttle. Replaced with a single `Map<id, update>` lookup → O(N+M).
3. **`fetchShuttles` could enter a fetch loop on PassioGO! REST/WS desync.** When the REST endpoint returned the `-1` "no buses" sentinel while the WS was still delivering frames, we'd `latestShuttles=[]` + `lastSeen.clear()`. The next WS frame is then "unknown" → triggers another on-demand `fetchShuttles` → REST still says `-1` → wipe again, hammering PassioGO! until they resync. Fix: if any `lastSeen` timestamp is within `STALE_TTL_MS` (2 min), trust the live WS feed and skip the wipe.
4. Doc comment fix: "Shuttle metadata refreshed hourly by cron" → "daily" (matches `CRON_MONITORS['fetch-transit'].schedule = '0 6 * * *'`).

All 12 service tests + 15 WS tests still pass. No type/lint regressions.

---

### Worth considering (not pushed)

**Backend**

- `passio-websocket.service.ts` — `this.shuttleGateway.broadcastShuttles(updates as unknown as Record<string, unknown>[])`. The `as unknown as` double cast is a smell. Easy fix: change `ShuttleTrackerGateway.broadcastShuttles` signature to `(shuttles: ShuttleLiveUpdate[])` (the gateway just `emit`s the array, so the type only documents intent — no runtime impact).
- `cron-monitors.ts` — `'fetch-transit'.maxRuntime: 60` (was 30). 60 minutes is generous for a job that does ~3 HTTP fetches; consider 5–10 min so Sentry actually catches a hang. Schedule change `0 0 * * *` → `0 6 * * *` is fine, but `docs/PRE_LAUNCH_AUDIT.md` line 244 still shows the old value — update if you want them in sync.
- `pruneStaleShuttles` doesn't write the pruned list to Redis. On Fly app restart we'll re-load the stale buses from Redis and only re-prune them 2 min later. Low priority since they get cleared on the next daily cron, but `await this.redis.set(REDIS_KEYS.SHUTTLES, this.latestShuttles, SHUTTLES_TTL_S)` after a non-zero prune would be consistent.
- `passio-websocket.service.ts` `hasActiveShuttles` changed from `Object.keys(data).length > 0` to `'busId' in data`. This is stricter and probably correct, but worth a sanity check that PassioGO! never sends a position frame keyed differently (e.g. older protocol versions).

**Mobile**

- `useTransitData.ts` — the socket effect's deps changed to `[]`. The previous version of the hook also inserted a placeholder for unknown bus IDs received via WS so the bus rendered immediately; that path was removed. With the backend fix in c883949 the bus now lands in `latestShuttles` server-side immediately, but mobile only sees it on the next `loadInitialShuttles()` poll (60s). Acceptable, but a 1–2s "manual GET on unknown id" could close the gap if you want instant rendering back.
- `useNearbyStopETAs.ts` — only the `isMounted` flag is checked before writing into `etaMap`. If `nearbyKey` changes mid-poll (user pans / lot changes), the stale fetchAll still resolves and pollutes `etaMap` with old stops' arrivals. Add a per-effect `let cancelled = false;` and `if (cancelled) return;` before each `setEtaMap` to scope the guard to the active effect run.
- `MapScreen.tsx`:
  - `routeColorMap = new Map(routes?.map(r => [r.id, r.color]) ?? [])` is rebuilt every render and is read inside the stops `.map()` (so every render rebuilds N maps). Wrap in `useMemo(..., [routes])`.
  - `handleRegionChangeComplete` calls `setMapBearing` on every region settle, which re-renders all `ShuttleMarker`s (each runs the heading subtraction). Throttling to ~150ms or only updating when the bearing delta exceeds a threshold (e.g. 1°) would cut this dramatically.
  - AsyncStorage hydration: `JSON.parse(lots[1])` / `JSON.parse(routes[1])` are not wrapped in try/catch. A corrupted entry would throw and the whole effect would silently fail (the outer `.catch(() => {})` swallows it). Wrap each parse individually.
  - Same effect: between mount and AsyncStorage resolving, `selectedLots=[]` so all lots render briefly, then snap to the persisted subset on resolve. A `hydrated` boolean (don't render the map content until hydration completes) avoids the flash.
- `transitProximity.ts` — `nearbyStopsForLot` measures distance from lot center to each polygon **vertex only**. That works for compact lots but for long/thin polygons (or a stop sitting near an edge midpoint between two distant vertices) you'll over-report distance. Consider point-to-segment distance, or at minimum document the vertex-only approximation.
- `ShuttleMarker.tsx` — `coordinate.timing({ toValue: 0, ... })` with the comment "unused at runtime" is fine for `react-native-maps` `AnimatedRegion.timing`, but worth a `// eslint-disable-next-line` or explanatory comment near the type assertion so future readers don't try to "clean it up."

**General**

- 24 of 40 changed files are touched without test updates (filter modal AsyncStorage round-trip, MapScreen polygon rendering, hexWithAlpha helper, centroid helper). Not blocking, but the polygon path is now the primary rendering branch and has no coverage.

LGTM after the small judgment calls above are addressed. Nice work on the batching + ping keep-alive — those tests in particular are great.

PR #232, opened 2026-05-10
Implement parked-lot UX, carpool detection, and push test toolkit

chuckmilton:
## Summary
This PR ships the full mobile parking/carpool/dev-toolkit upgrade plus backend support for safer push-notification testing.

### User-facing changes
- Added parked-lot highlighting on map with a distinct non-occupancy color.
- Added Find my car banner/action that recenters the map to the parked lot.
- Improved lot label readability at zoom by enforcing stable label sizing.
- Added carpool assistance UI (passenger mode, rider count, detection result card).

### Detection and state-machine changes
- Implemented carpool detection service with multi-signal inference.
- Integrated carpool detection lifecycle into EnhancedGeofencingProvider.
- Added persisted passenger mode/rider count and guarded in-session passenger toggles.
- Added realtime parking validation polling while in-lot.
- Added explicit INSUFFICIENT_DATA status for early active validation sessions.
- Preserved occupancy integrity: leave intent does not clear parked state; vehicular EXIT does.

### Dev/demo tooling improvements
- Replaced raw debug controls with scenario-based parking and leave-detection flows.
- Added test activity/geofence simulation hooks.
- Suppressed backend occupancy writes for test geofence events to avoid false 401/noise.
- Added local foreground push simulation controls in Dev Demo Toolkit.
- Added backend dev push-test endpoint for real backend → FCM send testing.
- Added push-test diagnostics (sent/pushConfigured/tokenCount) surfaced in app alerts.

### Security and release-hardening
- Debug push endpoint is disabled in production.
- Debug push endpoint now also requires ENABLE_DEBUG_PUSH_TEST=true.
- Push-token handling remains authenticated/scoped to current user.

## Tests and validation
- pnpm --filter mobile typecheck
- pnpm --filter backend typecheck
- pnpm --filter backend test -- src/notifications
- Added backend unit tests for debug push endpoint gating and diagnostics behavior:
  - notifications.controller.spec.ts
  - notifications.service.spec.ts

## Notes for reviewers
- Verify parked highlight/find-my-car behavior on iOS and Android maps.
- Verify carpool baseline and action behavior around ENTER/DWELL/EXIT transitions.
- Verify Dev Demo Toolkit local-sim and real-remote push flows (with ENABLE_DEBUG_PUSH_TEST=true).
- Confirm branch protections require ci + ios-build + android-build.


apps/backend/src/notifications/notifications.service.ts
```diff
@@ -56,6 +57,75 @@ export class NotificationsService implements OnModuleInit {
     await this.registerPushToken(user.id, token, platform);
   }
 
+  async sendDebugPushByEmail(email: string, dto: DebugSendPushDto): Promise<boolean> {
+    const user = await this.prisma.user.findUniqueOrThrow({
+      where: { email },
+      select: { id: true },
+    });
```
zachfpadilla, line 64: Is this dead code? I don't see it being referenced anywhere - it can be removed if not

apps/backend/src/notifications/notifications.controller.ts
```diff
@@ -56,5 +58,28 @@ export class NotificationsController {
...
+  @Post('me/push-test')
+  @HttpCode(HttpStatus.OK)
+  async sendPushTest(
+    @Req() req: AuthenticatedRequest,
+    @Body() dto: DebugSendPushDto,
+  ): Promise<{ sent: boolean; pushConfigured: boolean; tokenCount: number }> {
+    const debugPushEnabled = process.env.ENABLE_DEBUG_PUSH_TEST === 'true';
+    if (process.env.NODE_ENV === 'production' || !debugPushEnabled) {
+      throw new ForbiddenException('Push test endpoint is disabled (set ENABLE_DEBUG_PUSH_TEST=true in non-production environments)');
+    }
```
zachfpadilla, line 75: I feel like we should pull from `ConfigService` rather than directly from `.env`, which can be done like so:

```typescript
constructor(
  private readonly notificationsService: NotificationsService,
  private readonly configService: ConfigService,
) {}

const isProduction = this.configService.get<string>('NODE_ENV') === 'production';
const debugPushEnabled = this.configService.get<string>('ENABLE_DEBUG_PUSH_TEST') === 'true';
if (isProduction || !debugPushEnabled) {
  throw new ForbiddenException('...');
}
```

This would also mean the spec needs to be updated in the same way, but would make the tests both cleaner & more cohesive with the rest of the backend.

apps/backend/src/notifications/notifications.service.ts
```diff
@@ -56,6 +57,75 @@ export class NotificationsService implements OnModuleInit {
...
+        return {
+          title: 'Favorite Lot Clearing Up',
+          body: `${lotId} dropped below 30% occupancy.`,
+          data: { type: 'favorites_clearing', lotId },
+        };
+      case DebugPushType.SURGE:
+        return {
+          title: 'Campus Surge Alert',
+          body: 'Multiple lots are over 90% full right now.',
+          data: { type: 'surge', lotId },
```
zachfpadilla, line 113: As surge is campus-wide, the `SURGE` case should probably skip `lotId` in the same way that the `EVENTS` case does.

.github/workflows/mobile-test-hygiene.yml
```diff
@@ -0,0 +1,36 @@
+name: Mobile Test Hygiene
+
+permissions:
+  contents: read
+
+on:
+  schedule:
+    # 11:00 UTC daily
+    - cron: '0 11 * * *'
+  workflow_dispatch:
```
zachfpadilla, line 10: Consider adding a `pull_request` trigger here so that any `open-handle` issues are caught immediately (in conjunction with the `ci.yml` comment):

```yml
on:
  pull_request:
  schedule:
    # 11:00 UTC daily
    - cron: '0 11 * * *'
  workflow_dispatch:
```

.github/workflows/ci.yml
zachfpadilla, line 145: In conjunction with the `mobile-test-hygiene.yml` comment, adding a 15-minute limit would ensure that any hanging test jobs don't persist for the full 30 allotted minutes:

```yml
      - name: Run unit tests
        timeout-minutes: 15
        run: pnpm test -- --coverage
```

apps/mobile/__tests__/EnhancedGeofencingProvider.test.tsx
```diff
@@ -0,0 +1,342 @@
...
+  await act(async () => {
+    await Promise.resolve();
+  });
+}
+
+async function waitForProviderReady(): Promise<void> {
+  for (let i = 0; i < 6; i += 1) {
+    await flushEffects();
+  }
+}
```
zachfpadilla, line 184: nit: Instead of checking by 6*2 async ticks (which could theoretically fluctuate dependent on provider init time), it may be best to instead poll until `getAllLots` is called & throw an error if it isn't:

```typescript
async function waitForProviderReady(): Promise<void> {
  for (let i = 0; i < 20; i += 1) {
    await flushEffects();
    if (mockLotsApi.getAllLots.mock.calls.length > 0) return;
  }
  throw new Error('waitForProviderReady: provider did not complete initialization');
}
```

lawrenceDegoma (approved):
Everything on app works well. I'm liking the dev toolkit. The find my feature is also looking clean. Can't wait to test this physically as everything's looking great on paper lol

zachfpadilla (approved):
I see you've already merged, but I left a couple of nits anyways - nothing major though. Like @lawrenceDegoma said, really cool stuff! LGTM

chuckmilton:
Thank you for the thoughtful feedback @lawrenceDegoma and @zachfpadilla! I made the following changes based on the review comments. Much appreciated #GoSharks!!!

PR #236, opened 2026-05-11
fix(android): lot, logout, and geofence init fixes

zachfpadilla:
# Android: Lot Label, Logout, & Geofence Init Fixes

## Summary

Resolves several Android-specific bugs: lot labels were rendering strangely (very small text), a logout flow that misread a Chrome Custom Tab dismissal as user cancellation, a geolocation SDK re-init crash on app startup, and a clipped lot label caused by a `maxWidth`/`minWidth` conflict. Also fixes the geofencing status badge in ProfileScreen, which used hardcoded light-mode green values that clashed with the dark card surface.

## Changes

**`AzureAuth.tsx`** — Fixed Android logout being incorrectly treated as user cancellation. Chrome Custom Tabs always throws on end-session (even after a successful logout), so the Android branch is now evaluated first, before the generic "cancel" string check that was intercepting it.

**`locationService.ts`** — Fixed `"Waiting for previous start action to complete"` crash on app init. With `startOnBoot: true` and `stopOnTerminate: false`, the Background Geolocation SDK auto-restores its running state on Android before the JS side fires. `startGeofenceMonitoring()` now reads `getState()` first and skips `startGeofences()` if the SDK is already enabled.

**`MapScreen.tsx`** — Fixed lot label text from rendering too small regardless of component attributes due to bitmap rendering on Android skipping multi-pass layouts. For Android, `adjustsFontSizeToFit` is replaced with a custom `androidFontSize` selection.

**`ProfileScreen.tsx`** — Fixed geofencing status badge and "Currently in parking lot" info strip rendering with harsh light-green pastel backgrounds in dark mode. Both now use the same `rgba` tint + brighter foreground pattern already established by the advisory palette in `LotAmenities`.

**`theme.ts`** — Renamed `xxs2` → `xxxs` and added a true `xxxs` (8px) size to complete the font scale used by the Android lot label sizing logic.

## Screenshots

| iOS MapScreen | Android MapScreen |
|:-:|:-:|
| Screenshot 2026-05-10 at 7 47 59 PM | image |

### DIFF
--- a/apps/mobile/src/auth/AzureAuth.tsx
+++ b/apps/mobile/src/auth/AzureAuth.tsx
@@ -214,19 +214,19 @@ export const logoutFromAzure = async (idToken?: string): Promise<void> => {
       log('[AzureAuth] Azure AD logout successful');
     } catch (logoutError) {
       const errorMessage = (logoutError as Error).message || '';
-      
-      // Check if user explicitly cancelled the logout
-      if (errorMessage.includes('User cancelled') || errorMessage.includes('cancel')) {
-        log('[AzureAuth] User cancelled logout');
-        userCancelled = true;
+
+      if (Platform.OS === 'android') {
+        // Android Chrome Custom Tabs always throws on end-session (even on success);
+        // the server-side session is still invalidated — continue to clear local state.
+        log('[AzureAuth] Android logout error (continuing with local cleanup):', errorMessage);
       } else if (errorMessage.includes('error -3') || errorMessage.includes('org.openid.appauth.general')) {
         // AppAuth error -3 (OIDErrorCodeUserCanceledAuthorizationFlow) is expected with Azure AD
         // on iOS. The logout endpoint clears the session but doesn't redirect back properly.
         log('[AzureAuth] Azure AD session cleared (browser dismissed - this is expected)');
-      } else if (Platform.OS === 'android') {
-        // Android Chrome Custom Tabs may throw on end-session; the server-side
-        // session is still invalidated — continue to clear local state.
-        log('[AzureAuth] Android logout error (continuing with local cleanup):', errorMessage);
+      } else if (errorMessage.includes('User cancelled') || errorMessage.includes('cancel')) {
+        // iOS only: user explicitly dismissed the browser before completing logout
+        log('[AzureAuth] User cancelled logout');
+        userCancelled = true;
       } else {
         if (__DEV__) {
           console.warn('[AzureAuth] Azure AD logout encountered an issue, clearing local state:', logoutError);
--- a/apps/mobile/src/constants/theme.ts
+++ b/apps/mobile/src/constants/theme.ts
@@ -73,8 +73,8 @@ export const TYPOGRAPHY = {
     bold: 'Inter-Bold',
   },
   fontSize: {
+    xxxs: 8,
     xxs: 9,
-    xxs2: 11,
     xs: 10,
     sm: 12,
     md: 14,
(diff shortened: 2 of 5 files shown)


chuckmilton (approved):
Great work on making the fixes! Everything LGTM with a couple of small observations, none blocking though.

### `apps/mobile/src/auth/AzureAuth.tsx`
Reordering the Android branch ahead of the `"User cancelled"` string match is the right call. Chrome Custom Tabs always throws on `endSession` (even on success), and the Android RN runtime localizes / reformats the error message, so the substring check was a false-positive trap. Treating any logout error on Android as "session cleared on the device" matches what `react-native-app-auth` actually delivers.

Side effect worth knowing: we can no longer detect a real user-cancel on Android — but for logout that's the desired behavior (always clear local state). Good.

### `apps/mobile/src/constants/theme.ts` (non-blocking)
`xxs2: 11` → `xxxs: 8` is technically two changes at once (rename **and** value change). Confirmed via `git grep "xxs2" apps/mobile/src/**` that no live source consumes the old key (only stale `android/app/build/**` sourcemaps and `coverage/**` artifacts). Safe.

Nit: the resulting map is now `xxxs:8, xxs:9, xs:10, sm:12, …` — order in the object literal doesn't matter, but if you want to keep the file readable consider re-sorting the keys ascending so future contributors don't have to think.

### `apps/mobile/src/screens/MapScreen.tsx`
Smart fix for the Android map-marker bitmap-cache problem (Marker Text doesn't get multi-pass layout, so `adjustsFontSizeToFit` silently no-ops). Validated the threshold ladder against actual `lot_name` values:

| Length bucket | Font | Lot examples |
|---|---|---|
| ≤12 | `xs` (10) | `Lot G1`–`Lot G14`, `Lot E1`–`Lot E11` (6–7 chars) |
| ≤16 | `xxs` (9) | `Palo Verde North`, `Palo Verde South` (16 chars) |
| else | `xxxs` (8) | `Pyramid Parking Structure` (25 chars) |

`maxWidth: 56` + `numberOfLines={3}` at 8px Inter fits "Pyramid / Parking / Structure" cleanly. 

Two small things:
1. The base `lotText` style still has its own `fontSize` (looking at line 928+); when Android overrides via the inline `{ fontSize: androidFontSize }` it works because of style-array precedence, but it'd be one less mental hop to gate the base font on iOS only or just drop the base size since both branches now provide one.
2. iOS still uses `adjustsFontSizeToFit` + `numberOfLines={3}`, which means single-word lots like "Pyramid Parking Structure" can now wrap to 3 lines on iOS too instead of shrinking to one. That's almost certainly fine (and probably more readable) but worth a quick visual confirmation on an iOS build before merging.

### `apps/mobile/src/screens/ProfileScreen.tsx`
Geofence status badge + "Currently in parking lot" strip now use `rgba(16,185,129,0.18)` / `#6ee7b7` in dark mode — matches the LotAmenities advisory palette. Clean fix, no concerns.

### `apps/mobile/src/services/locationService.ts`
Early-return on `getState().enabled` is the correct way to handle the "Waiting for previous start action to complete" race that `startOnBoot:true` introduces. The native plugin restores its prior config from disk on boot, so calling `start*()` before our JS-side init catches up will reject.

Minor note (not a request to change): `this.trackingMode = 'geofences'` after the early-return assumes the restored state is geofence mode. That's true for our config (we never call `.start()` for continuous tracking from a cold boot path), but if we ever add a "track on boot during active session" mode we'd want to read the actual mode out of the SDK or out of persistent state instead.

DottL (approved):
Lookie Goodie to me 🌻

